PR #11, opened 2021-05-11
Jupyter notebook coast

hcwinsemius:
- bug fixes in plot_basemap and in setup_gauges
- notebooks with examples


examples/sfinc_update_model.ipynb
```diff
@@ -0,0 +1,375 @@
...
+   "metadata": {},
+   "source": [
+    "### contents of HydroMT Sfincs configuration .ini file\n",
+    "We can provide many more settings in an ini file instead of command line arguments. These settings point to the data catalog in many cases. \n",
+    "\n",
+    "###################\n",
+    "DIRK, can I refer to a notebook explaining this in more detail?\n",
+    "###################\n",
+    "\n",
+    "If you don't provide an explicit ini file, a default .ini file is used for the model building process. Let's have a look at this. "
```
DirkEilander, line 54: >If you don't provide an explicit ini file, a default .ini file is used for the model building process. Let's have a look at this. 

This is no longer the case. By default only the basemaps with default settings are created. An ini-file is required to build other components.

examples/sfinc_update_model.ipynb
```diff
@@ -0,0 +1,375 @@
...
+   "outputs": [],
+   "source": [
+    "!hydromt --models"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### contents of HydroMT Sfincs configuration .ini file\n",
```
DirkEilander, line 47: I moved this part to the other notebook as I think it's nice to show how to build an entire model in that notebook.

hydromt_sfincs/sfincs.py
```diff
@@ -560,12 +560,14 @@ def setup_gauges(self, gauges_fn=None, **kwargs):
         """
         if gauges_fn is not None:
             name = self._GEOMS["gauges"]
+            # ensure the catalog is loaded before adding any new entries
+            self.data_catalog.sources
             gdf = self.data_catalog.get_geodataframe(
-                filename, geom=self.region, assert_gtype="Point"
+                str(gauges_fn), geom=self.region, assert_gtype="Point", **kwargs
```
DirkEilander, line 566: good catch!

The data_catalog.get_geodataframe method now also handles Path instances, see https://github.com/Deltares/hydromt/pull/26

examples/sfinc_update_model.ipynb
```diff
@@ -0,0 +1,375 @@
...
+    "If you don't provide an explicit ini file, a default .ini file is used for the model building process. Let's have a look at this. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fn_ini = '../hydromt_sfincs/data/sfincs/sfincs_build_default.ini'\n",
```
DirkEilander, line 63: I removed this file as it is not used by hydroMT. 

Also good to know: the notebooks can only contain references to files in the example folder. Other references are not found when building the docs based on the notebooks.

DirkEilander (commented):
Nice work Hessel! I will commit a few small changes and then merge.

PR #12, opened 2021-05-12
Setup rivers downstream

Leynse:
Addition of setup_rivers_downstream functionality and riverine case example plus notebook


DirkEilander:
Thanks for this PR @Leynse !

I've made a few changes to the PR before merging:
- renamed `setup_river` to `setup_river_inflow` and `setup_river_downstream` to `setup_river_outflow` 
- fixed some relative paths in the ini files
- removed gis/bat data from examples
- added riverine test

PR #16, opened 2021-05-12
Jupyter notebook coast

hcwinsemius:
- notebook for coast extended with alternative landarea mask
- fixed small bug in posixpath being treated as str in sfincs.py


DirkEilander:
This PR has been idle for a while and I believe is outdated.  Feel free to re-open if my judgement is wrong.

PR #32, opened 2021-08-06
update setup_mask

DirkEilander:
- removed default values and new option to combine a polygon and elevation based criteria after @Leynse suggestion.


Leynse (approved):
Seems to be a nice generic implementation of my changes

PR #48, opened 2022-01-25
Setup region

### LINKED ISSUE #43: region based on mask
Currently the region `property` is based on the bounding box of the model as defined in the general hydroMT `Model` class. We should overrule this for SFINCS to base region on the model mask. That makes it possible to call e.g. `setup_rivers_inflow` based on any existing model.

### LINKED ISSUE #44: re-introduce subbasin regions
Ideally in a seperate setup_regions method to allow using a DEM from a different source than the one used for determening the subbasins. this requires changes upstream in hydroMT.

### LINKED ISSUE #47: setup_h_forcing select locations based on water level boundary cells
instead of a buffer around the model domain, set the buffer around the water level boundary cells.

DirkEilander:
- new `setup_region` method to set the (hydrological) model region of interest (before part of `setup_basemaps`). fixes #44 
- as a consequence `setup_basemaps` has been replaced by `setup_topobathy` with topobathy_fn (instead of basemaps_fn argument) and example notebooks have been updated accordingly.
- In `setup_river_inflow` and `setup_river_outlfow` the model domain is determined on-the-fly, thus it is not required to run setup_mask beforehand. fixes #43 
- In `setup_h_forcing` the buffer is relative to mask 2 cell instead of all model region, fixes #47 
- bugfix in bathymetry.get_river_bathymetry with rivwth_method='geom' thanks to @JoostBuitink 
- `setup_mask`  has a "reset_mask" option to start with a clean mask (default True). Does not change current behaviour.
- `setup_mask` takes a new "drop_area" argument to drop regions of contiguous cells smaller than this maximum area threshold, useful to remove (spurious) small islands. Used to be "min_cells" argument and defined in number of cells instead of area.
- `setup_mask` takes a new "fill_area" argument to fill regions of contiguous cells below the "min_elv" or above "max_elv" threshold surrounded by cells within the valid elevation range. Used to be boolean "fill_holes" argument.
- In `setup_bounds` and `setup_mask` a "connectivity" argument is exposed to determine whether edge cells or regions of contiguous cells should be based on D4 (horizontal and vertical) or D8 (also diagonal) connections. Default is D8. This used to be D8 for bounds and D4 for mask.


Leynse (approved):
Thanks for further improving the code!

PR #50, opened 2022-01-28
improve basemap plot

DirkEilander:


Leynse:
Nice quick fix

Leynse (approved):
Thanks for fixing

PR #54, opened 2022-02-11
various mask and bounds improvements

DirkEilander:
- distinguish between `mask_fn` which provides a starting point for active / boundary cells and `include_mask_fn` and `exclude_mask_fn` which overrides any previous criteria in both `setup_bounds` and `setup_mask` methods.
- `setup_river_inflow` has a closed_bounds_buffer to make sure boundary cells within a radius from source points are closed.
- add some new switches in `plot_basemaps`

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -31,7 +31,7 @@ Changed (**Breaking**)
 
 Changed
 ^^^^^^^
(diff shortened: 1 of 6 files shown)


Leynse (approved):
Thanks a lot for merging the 2 setup_river_inflow functions!

1 suggestion, consider keep naming of buffer/river width the same betwen setup_river_inflow and _outflow functions.

e.g. 'closed_bounds_river_width' instead of 'closed_bounds_buffer' for setup_river_inflow

PR #56, opened 2022-02-21
Changing data_vars keys when reading sfincs_map.nc

couasnonanais:
Work around to change the naming of 'corner_' keys because this was giving an error in the newer version.


DirkEilander:
@couasnonanais Thanks for this PR. After giving it a bit more thought I made the changes in a way that it also accommodates for future changes in naming, see #57. Will close this PR.

PR #79, opened 2023-04-11
new river in/outlfow methods

DirkEilander:


roeldegoede (approved):
Great! Works perfect for our australia project

PR #90, opened 2023-05-02
added setup_drainage_structures to account for pumps/culverts

roeldegoede:
Since the format is quite different compared to weirs/thindams, I created a new method and some new read/write utils functions.


hydromt_sfincs/sfincs.py
```diff
@@ -1186,6 +1187,97 @@ def setup_structures(
...
+                        )
+                        gdf_structures = gdf_structures.drop(index=i)
+                    else:
+                        # convert Multilinestring to LineString
+                        gdf_structures.loc[i, "geometry"] = LineString(ls.coords)
+            elif isinstance(row.geometry, LineString) and len(row.geometry.coords) != 2:
+                self.logger.debug(
+                    f"Row {i} contains a LineString with {len(row.geometry.coords)} points."
+                )
+                gdf_structures = gdf_structures.drop(index=i)
```
DirkEilander, line 1267: Instead of dropping incorrect lines we could also just simplify each linestring based on its start and endpoints with the following code. Up to you.

```
# multi to single lines
lines = gdf_structures.explode(column="geometry").reset_index(drop=True)
# get start [0] and end [1] points
endpoints = lines.boundary.explode().unstack()
# merge start and end points into a single linestring
gdf_structures['geometry'] = endpoints.apply(lambda x: LineString(x.values.tolist()), axis=1)
```

If we keep the current approach this can be improved by avoiding looping over rows of pandas/geopandas object. This can be achieved with 

```
# multi to single lines
gdf = gdf_structures.explode(column="geometry").reset_index(drop=True)
# get number of points in line
ncoords = gdf.geometry.apply(lambda x: len(x.coords))
# drop lines with more than 2 points
if np.any(ncoords != 2):
    # add warning
    gdf = gdf[ncoords == 2]
```

hydromt_sfincs/utils.py
```diff
@@ -623,6 +625,97 @@ def read_geoms(fn: Union[str, Path]) -> List[Dict]:
...
+        "par4",
+        "par5",
+    ]
+
+    gdf = copy.deepcopy(gdf_drainage)
+    # get geometry linestring and convert to xsnk, ysnk, xsrc, ysrc
+    gdf["xsnk"] = [ls.coords[0][0] for ls in gdf.geometry]
+    gdf["ysnk"] = [ls.coords[0][1] for ls in gdf.geometry]
+    gdf["xsrc"] = [ls.coords[1][0] for ls in gdf.geometry]
+    gdf["ysrc"] = [ls.coords[1][1] for ls in gdf.geometry]
```
DirkEilander, line 660: Should the source be the start point of each line and the sink the endpoint? If so this should still be adjusted in my suggestion below. 

Here you can also avoid the loops using
```suggestion
    endpoints = gdf.boundary.explode().unstack()
    gdf["xsnk"] = endpoints[0].x
    gdf["ysnk"] = endpoints[0].y
    gdf["xsrc"] = endpoints[1].x
    gdf["ysrc"] = endpoints[1].y
```

docs/changelog.rst
```diff
@@ -12,6 +12,10 @@ Bugfix
 ------
 - bugfix in **write_forcing** to ensure all NetCDF files are written instead of only the first one.
 
+New
+---
+- **setup_drainage_structures** to setup drainage structures (pumps,culverts) from a geodataframe.
```
DirkEilander, line 17:
```suggestion
- **setup_drainage_structures** to setup drainage structures (pumps,culverts) from a geodataframe. PR#90
```

DirkEilander (commented):
Nice work @roeldegoede! I've made a couple of minor suggestions and two suggestion to avoid loops when working with geopandas.GeoDataFrames.

PR #91, opened 2023-05-03
add transpose to ensure y,x coordinates instead of x,y

roeldegoede:
Simple tranpose to ensure coordinates are always in the order y,x and corner_y,corner_x (instead of x,y) to support the sfincs_map of the Alpe d'Huez release (and earlier unreleased versions)

### DIFF
--- a/hydromt_sfincs/utils.py
+++ b/hydromt_sfincs/utils.py
@@ -673,11 +673,7 @@ def read_sfincs_map_results(
 
     # support for older sfincs_map.nc files
     # check if x,y dimensions are in the order y,x
-    if ds_map.xc.dims[0] != "y":
-        ds_map = ds_map.transpose(..., "y", "x")
-    # check if corner_x,corner_y dimensions are in the order corner_y,corner_x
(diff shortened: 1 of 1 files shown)


hydromt_sfincs/utils.py
```diff
@@ -671,6 +671,14 @@ def read_sfincs_map_results(
         [var for var in ds_map.data_vars.keys() if (var in rm.values())]
     )
 
+    # support for older sfincs_map.nc files
+    # check if x,y dimensions are in the order y,x
+    if ds_map.xc.dims[0] != "y":
+        ds_map = ds_map.transpose(..., "y", "x")
+    # check if corner_x,corner_y dimensions are in the order corner_y,corner_x
+    if ds_map.corner_xc.dims[0] != "corner_y":
+        ds_map = ds_map.transpose(..., "corner_y", "corner_x")
```
DirkEilander, line 680: If all these coordinates are always present (which should be the case right?) we could basically simplify this.

```suggestion
    ds_map = ds_map.transpose(..., "y", "x", "corner_y", "corner_x")
```

PR #95, opened 2023-06-02
94 bug memory error when loading large raster datasets clipped to a geometry

roeldegoede:


hydromt_sfincs/workflows/merge.py
```diff
@@ -88,9 +88,20 @@ def merge_multi_dataarrays(
...
-        da1 = da1.raster.reproject_like(da_like, method=method).load()
+        da1 = da1.raster.clip_bbox(
+            da_like.raster.box.to_crs(da1.raster.crs.to_epsg()).total_bounds, buffer=10
+        )
+        if np.any(np.array(da1.shape) <= 2):
+            # no data in da1 so use an empty array like da_like
+            logger.debug("No data da1, start with nan array")
+            da1 = xr.full_like(da_like, np.nan)
+        else:
+            da1 = da1.load()
```
roeldegoede, line 99: These load statement were necessary for the subgrid determination right? Because they are still sometimes causing memory issues

hydromt_sfincs/workflows/merge.py
```diff
@@ -88,9 +88,20 @@ def merge_multi_dataarrays(
...
+            da1 = da1.raster.reproject_like(da_like)
     elif reproj_kwargs:
-        da1 = da1.raster.reproject(method=method, **reproj_kwargs).load()
+        #TODO also clip to bbox?
```
roeldegoede, line 102: I don;t think this option is used in the code anymore .. However, if we want to koop it, maybe we have to add the clipping to the bbox here as well?

hydromt_sfincs/sfincs.py
```diff
@@ -2886,11 +2899,25 @@ def _parse_datasets_dep(self, datasets_dep, res):
...
+                    if da_elv.raster.crs.is_compound:
+                        da_elv.raster.set_crs(da_elv.raster.crs.sub_crs_list[0])
+
+                    # cropping went wrong for very large datasets with crs != 4326 due to cropping with geom
+                    da_elv = da_elv.raster.clip_bbox(
+                        self.mask.raster.box.to_crs(
+                            da_elv.raster.crs.to_epsg()
+                        ).total_bounds,
+                        buffer=10,
+                    )
```
DirkEilander, line 2919: Instead of this fix here wouldn't it be better to change the data itself? This is easily done for geotifs with `rio` a command line tool that comes with rasterio, e.g., `rio edit-info --crs EPSG:3857 example.tif` (this also works for vrt or other gdal raster files)

We can also deal with compound crs in the core, but this needs a bit more thinking. Can you make an issue there?

hydromt_sfincs/workflows/merge.py
```diff
@@ -88,9 +88,20 @@ def merge_multi_dataarrays(
         method = "bilinear"
 
     if da_like is not None:  # reproject first raster to destination grid
-        da1 = da1.raster.reproject_like(da_like, method=method).load()
+        da1 = da1.raster.clip_bbox(
+            da_like.raster.box.to_crs(da1.raster.crs.to_epsg()).total_bounds, buffer=10
+        )
```
DirkEilander, line 93: Is this clipping required before reprojecting? If so, it would be good to add this in the reproject_like method in the core. In my previous tests it didn't seem to matter.

hydromt_sfincs/workflows/merge.py
```diff
@@ -115,6 +126,13 @@ def merge_multi_dataarrays(
         # base reprojection method on resolution of datasets
         reproj_method = da_list[i].get("reproj_method", None)
         da2 = da_list[i].get("da")
+        da2 = da2.raster.clip_bbox(
+            da1.raster.box.to_crs(da2.raster.crs.to_epsg()).total_bounds, buffer=10
+        )
+        if np.any(np.array(da2.shape) <= 2):
+            print(f"No data for this tile")
```
DirkEilander, line 133: replace print statements by logger

hydromt_sfincs/workflows/merge.py
```diff
@@ -209,11 +227,9 @@ def merge_dataarrays(
...
     ## reproject da2 and reset nodata value to match da1 nodata
     try:
-        da2 = (
-            da2.raster.reproject_like(da1, method=reproj_method)
-            .raster.mask_nodata()
-            .load()
-        )
+        da2 = da2.load()
+        da2 = da2.raster.reproject_like(da1, method=reproj_method)
+        da2 = da2.raster.mask_nodata()
```
DirkEilander, line 232: Before this was reprojecting first and then loading. I would only load after reprojecting to make sure you don't load the entire grid. I noticed you added a clip before calling this function, but in case it's called elsewhere this may cause memory issues.

DirkEilander (commented):
Nice work @roeldegoede. The PR does more than just replacing the clip_geom instances. It would be good to discuss the changes in subgrid.py and merge.py, why these are necessary and if this should (also) lead to changes in the core.

hydromt_sfincs/sfincs.py
```diff
@@ -2886,11 +2899,22 @@ def _parse_datasets_dep(self, datasets_dep, res):
...
                         dataset.get("elevtn", dataset.get("da")),
-                        geom=self.mask.raster.box,
-                        buffer=10,
+                        # bbox=self.mask.raster.box.to_crs(4326).total_bounds,
+                        # buffer=10,
                         variables=["elevtn"],
                         zoom_level=(res, "meter"),
                     )
+
+                    #TODO properly fix this in hydromt core ?
```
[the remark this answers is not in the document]
DirkEilander: See my comment below. We can also think about a fix in the core.

hydromt_sfincs/workflows/merge.py
DirkEilander: Not sure. Would be good to take some time to profile these function to understand what works best, which load / clip statements are required. My understanding is that the way the data is stored is also very important to avoid memory errors. Perhaps we should plan an afternoon to work on this.

hydromt_sfincs/workflows/merge.py
DirkEilander: see also comment below

DirkEilander (approved):

PR #101, opened 2023-06-20
ENH: added setup_storage_volume to account for green-infratructure

roeldegoede:
added an additional layer to the SfincsModel.grid to account for green-infrastructure. This layer describes a storage volume in m3 per cell that needs to be filled first before the cell becomes wet, and therefore mimics the presence of green-infrastructure.

input can be 
- a polygon with a volume in m3 that is divided over the cells within the polygon
- a polygon with a height in m, then internally the volume is computed by multiplying the area and the hegiht
- points with a volume


hydromt_sfincs/sfincs.py
```diff
@@ -1572,6 +1573,130 @@ def setup_drainage_structures(
...
+                    x, y = single_gdf.geometry.iloc[0].coords[0]
+
+                    # check if the grid is rotated
+                    if da_vol.raster.rotation != 0:
+                        # rotate the point
+                        x, y = ~da_vol.raster.transform * (x, y)
+                        # select the grid cell nearest to the point
+                        closest_point = da_vol.reindex(x=da_vol.x, y=da_vol.y).sel(
+                            x=x, y=y, method="nearest"
+                        )
```
DirkEilander, line 1661: I think this could be replaced with `row,col = da_vol.raster.rowcol(x, y)` after which you can use da_vol.iloc[row, col] later?

hydromt_sfincs/sfincs.py
```diff
@@ -1572,6 +1573,130 @@ def setup_drainage_structures(
...
+                    else:
+                        # select the grid cell nearest to the point
+                        closest_point = da_vol.sel(x=x, y=y, method="nearest")
+
+                    # add the volume to the grid cell
+                    if single_vol is not None and not np.isnan(single_vol).any():
```
DirkEilander, line 1667: strictly speaking I don't think the any() is required here because single_vol is always a float right? Idem in a couple of instances below and for single_height.

DirkEilander (changes requested):
I've just checked the code, not actually tested it. 
Apart from some small comments, before merging, it would be good to 
- move the lines 1611-1691 to a workflow script
- add a unit test with a gdf containing a point a polygon with and without volume/height attributes on a small rotated grid to make sure it works as expected.

hydromt_sfincs/sfincs.py
roeldegoede: This doesn't seem to work for me ..

DirkEilander (approved):
nice work on the test Roel, looks good to me!

PR #103, opened 2023-07-05
Create constant infiltration rates based on lulc and reclass_table

roeldegoede:
In addition to the existing functionality, we can now also set up constant infiltration rates based on landuse/landcover and a reclass table. Before we could only use spatiallly varying infiltration fields.


DirkEilander (approved):
Thanks @roeldegoede for this PR!
I've made a few small changes (see also comments), let me know if you agree.
It would be good to also update the csv tables - can you still do that?
Otherwise good to be merged.

PR #104, opened 2023-07-05
enabled to provide only locations for setup_waterlevel_forcing

### LINKED ISSUE #97: discharge inputs using mask
Add a new function or update the setup_discharge_forcing to allow the user to select a subset of discharge inputs from a dataset with a larger number of points (e.g., netcdf). Basically add the same functionality that the setup_waterlevel_forcing has. Example of my work around is attached...

Screenshot 2023-06-15 at 1 27 35 PM

roeldegoede:
- Allow locations only in "SfincsModel.setup_waterlevel_forcing" and "SfincsModel.setup_discharge_forcing"
- Add buffer in "SfincsModel.setup_discharge_forcing" fixes #97

### DIFF
--- a/hydromt_sfincs/sfincs.py
+++ b/hydromt_sfincs/sfincs.py
@@ -1597,8 +1597,6 @@ class SfincsModel(GridModel):
                 index_col=0,
             )
             df_ts.columns = df_ts.columns.map(int)  # parse column names to integers
-        else:
-            raise ValueError("Either geodataset or timeseries must be provided")
 
         # optionally read location data (if not already read from geodataset)
         if gdf_locs is None and locations is not None:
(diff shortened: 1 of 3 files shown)


hydromt_sfincs/utils.py
```diff
@@ -317,7 +317,7 @@ def write_timeseries(
     data[:, 0] = (df.index - tref).total_seconds()
     # calculate required width for time column; hard coded single decimal precision
     # format for other columns is based on fmt`argument
-    w = int(np.floor(np.log10(data[-1, 0]))) + 3
+    w = int(np.floor(np.log10(abs(data[-1, 0])))) + 3
```
DirkEilander, line 320: Accidently found this "bug". This line errors if tref > tstart. This shouldn't be the case normally and not sure how SFINCS deals with it. With the abs statement it always works.

hydromt_sfincs/sfincs.py
```diff
@@ -1625,13 +1628,7 @@ def setup_waterlevel_forcing(
             )
 
         # set/ update forcing
-        if df_ts is not None:
-            # set timeseries and locations
-            self.set_forcing_1d(df_ts, gdf_locs, name="bzs", merge=merge)
-        else:
-            # set locations only
-            self.set_forcing_1d(gdf_locs=gdf_locs, name="bzs", merge=merge)
```
DirkEilander: The if-else isn't necessary as set_forcing also works if either df_ts or gdf_locs is None

DirkEilander (approved):
I've made a few small changes. If you agree this can be merged.

PR #110, opened 2023-07-24
Fix replacing the mask values with one (due to dilating) by making a …

roeldegoede:
## Issue addressed

@taiwoogunwumi found out that the mask seems to be resetted after setup_subgrid

## Explanation

While generating the subgrid with setup_subgrid, ndimage.binary_dilation not only replaces the values of da_mask_block but also the ones of da_mask. Therefore, setup_subgrid basically resets the boundary mask cells. This is not desired.

By making a copy, the problem seems te be solved.


DirkEilander (approved):
Good catch! I've just reformatted the code for readability.

PR #112, opened 2023-07-24
more robust alignment of grid origin

DirkEilander:
in SfincsModel.setup_grid_from_region replace `dec_origin` argument by `align_origin_to_res`. The dec_origin default value was very prone to issues especially when working with geographic CRS.


DirkEilander:
superseded by #138

PR #114, opened 2023-07-28
add setup_observation_lines including example and test

roeldegoede:
Added functionality to read, write and create crs (cross-section) files to the SfincsModel. Through these cross-sections, discharges are monitored.

In SFINCS, the LineStrings are snapped to all u- and v points for which all the discharged are added together. Discharges per cross-section are added to the sfincs_his.nc.


hydromt_sfincs/sfincs.py
```diff
@@ -1316,6 +1317,49 @@ def setup_observation_points(
...
+        ---------
+        locations: str, Path, gpd.GeoDataFrame, optional
+            Path, data source name, or geopandas object for observation lines (cross-sections).
+        merge: bool, optional
+            If True, merge the new observation lines with the existing ones. By default True.
+        """
+        name = self._GEOMS["observation_lines"]
+
+        # FIXME ensure the catalog is loaded before adding any new entries
+        self.data_catalog.sources
```
DirkEilander, line 1342: Why is this needed?

hydromt_sfincs/sfincs.py
```diff
@@ -1316,6 +1317,49 @@ def setup_observation_points(
...
+
+        # FIXME assert_gtype="LineString" does not work for MultiLineString and default seems to be Point (??)
+        gdf_obs = self.data_catalog.get_geodataframe(
+            locations, geom=self.region, assert_gtype=None, **kwargs
+        ).to_crs(self.crs)
+
+        # make sure MultiLineString are converted to LineString
+        gdf_obs = gdf_obs.explode().reset_index(drop=True)
```
DirkEilander, line 1350: The original data indices are ignored here. Is it important that the indices of the original data are kept? I would think that it might be useful to match indices in the results to your input data. I noticed that also in obs the indices get 'lost' when reading or merging. 

Not necessarily something we need to solve here and now, but perhaps good to discuss in conjunction with #115

hydromt_sfincs/plots.py
```diff
@@ -19,6 +19,7 @@
     "bnd": dict(marker="^", markersize=75, c="w", edgecolor="k", annotate=True),
     "src": dict(marker=">", markersize=75, c="w", edgecolor="k", annotate=True),
     "obs": dict(marker="d", markersize=75, c="w", edgecolor="r", annotate=True),
+    "crs": dict(linestyle=":", linewidth=1.0, color="k", annotate=False),
```
DirkEilander, line 22: this is a bit difficult to see with the dep file in the background. Perhaps use red color?

DirkEilander (approved):
Very nice PR @roeldegoede! Just added to points for discussion, but the PR can be merged as is I think.

hydromt_sfincs/sfincs.py
roeldegoede: Good question, copied this from the setup_observation_points. Can this be related to the fact that self.data_catalog normally keeps tracks of all the used data_sources (something which seems to be disappeared along the way?) and therefore we needed to ensure that the data_catalog was loaded?

Maybe this still originates from the first codes and the problem has been fixed in the meanwhile?

hydromt_sfincs/plots.py
roeldegoede: Agree, do you want to stick to the ":" linestyle or are there also better options (exceptthe solid line I guess)? I often notice that dashed or dotted lines do not cover the entire line from start to end, because that may coincide with the "white spaces" in between the icons.

hydromt_sfincs/sfincs.py
roeldegoede: Fair point, since we give the observation points and the observation_lines a name, most of the times that would be sufficient. But maybe keeping the original indices would be nice indeed!

PR #119, opened 2023-08-08
fix constant offset in dataset_dep

DirkEilander:
A constant offset value in dataset_dep is currently ignored. This small PR fixes the issue.


DirkEilander:
Since it's a very small PR and the breaking tests are based on HydroMT-core ([issue solved in this PR](https://github.com/Deltares/hydromt/pull/473)) I've decided to go ahead an merge it to avoid building wrong models in our project.

PR #120, opened 2023-08-09
Some improvements to the plot_basemaps method

DirkEilander:
This PR 
- adds more background maps support by contextily
- makes it possible to plot other than "grid" variables on the basemap, e.g. the hmax flood map
- fixes a bug that ignores the write_gis argument when writing forcing bnd and src files
-


roeldegoede:
Is this ready for review, or are there more improvements that you want to add? Maybe we can try to tackle #123 in this PR as well?

DirkEilander:
> Is this ready for review, or are there more improvements that you want to add? Maybe we can try to tackle #123 in this PR as well?
I don't have an easy dataset to test that issue with available here. I suggest to fix that separately. I'll add a comment in that issue on a possible solution direction.

roeldegoede (approved):
Nice work and very clear explanations. Everything seems to work on my side as well, so ready for merging!

PR #121, opened 2023-08-11
Fix tests

DirkEilander:
Since [HydroMT PR473](https://github.com/Deltares/hydromt/pull/473) we cannot use private model components in our code (this should already not have been the case). E.g., we should always use `Model.geoms` instead of `Model._geoms`. While fixing this I discovered a few other issues which are solved here:

- in read_states we tried to read the index file while this is not necessary when we have a mask file already set.
- the test_model_build was still using staticmaps and staticgeoms which have long been deprecated and hence not actually testing the model. 
- test against hydromt release and latest dev version. This will help to understand what causes an issue.


DirkEilander:
TODO
- [ ] add check on the `SfincsModel.subgrid` component in `test_model_build`

roeldegoede (approved):

PR #124, opened 2023-09-08
Update sfincs.py

Leynse:
- Xarray.rename input type seems to have changed
- Only relevant for precip_2D and press_2D? (or more?)

### DIFF
--- a/hydromt_sfincs/sfincs.py
+++ b/hydromt_sfincs/sfincs.py
@@ -2095,8 +2095,8 @@ class SfincsModel(GridModel):
                     downsampling="sum",
                     logger=self.logger,
                 )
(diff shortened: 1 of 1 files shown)


DirkEilander:
As discussed with @Leynse via chat, I think this issue is caused by the fact that the method is called with a `xarray.Dataset` instead of `xarray.DataArray` object for the precip and press arguments. These arguments were not clipped or filtered in the released HydroMT `DataCatalog.get_rasterdataset` functionality, but this has been solved in https://github.com/Deltares/hydromt/pull/481. Now when given a dataset and then parsing it with e.g. `DataCatalog.get_rasterdataset(variables=['precip']` it should only return the precip variable as DataArray, Hence I think this fix is not necessary.

PR #128, opened 2023-09-13
don't clip geotiffs to region, but only mask inactive cells

roeldegoede:
Because of clipping of the rasterdata to the region while writing away the geotiffs to the gis-folder, there was a mismatch between the size of the gis-data and the model grid data. This may cause issues while reading the model.

This problem somehow only arised for rotated models, and also only when the gis-data is used, which is the case when the depfile is not written because of the presence of the subgridfile.


TBovenschen (approved):
Looks good

PR #138, opened 2023-10-20
align grid with resolution

roeldegoede:
For certain projects, it is desirable to align the SFINCS grid with already existing grids. This can be easily achieved by rounding of the origin of the grid.

The proposed changes are in line with already existing functionality in HydroMT-core [(see here)](https://github.com/Deltares/hydromt/blob/31d69df14e5e42e0948f533e82cbabb269f3faee/hydromt/models/model_grid.py#L458) 

At the moment HydroMT-SFINCS makes distinction between setup_grid and setup_grid_from_region (the latter is in line with the setup_grid from the core), which makes it hard to use the core functionality. I propose to have it temporary fixed like this and find a more integral solution in the future to prevent double maintenance work.

### DIFF
--- a/hydromt_sfincs/sfincs.py
+++ b/hydromt_sfincs/sfincs.py
@@ -259,6 +259,7 @@ class SfincsModel(GridModel):
             based on a (sub)(inter)basins without a 'bounds' argument.
         align : bool, optional
             If True (default), align target transform to resolution.
+            Note that this has only been implemented for non-rotated grids.
         dec_origin : int, optional
             number of decimals to round the origin coordinates, by default 0
         dec_rotation : int, optional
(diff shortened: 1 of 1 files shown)


hydromt_sfincs/sfincs.py
```diff
@@ -256,6 +257,8 @@ def setup_grid_from_region(
             Name of data source with basin (bounding box) geometries associated with
             the 'basins' layer of `hydrography_fn`. Only required if the `region` is
             based on a (sub)(inter)basins without a 'bounds' argument.
+        align : bool, optional
+            If True (default), align target transform to resolution.
```
DirkEilander, line 261: This only works for non-rotated right?

```suggestion
            If True (default), align target transform to resolution.
            Note that this has only been implemented for non-rotated grids.
```

DirkEilander (approved):
Apart from a small note it the docstring this looks good to me.

PR #141, opened 2023-10-30
region.geojson changed, whereas all other files remained equal ..

roeldegoede:
## Problem 
Tests were failing on assert_geodataframe_equal (Line 293 in test_model_build) for the region. Visually the regions were still equal, but somehow the decimals changed in the generation of the region...

## Analysis
Not sure where this is introduced, but maybe different versions of some packages result in slightly different geometries?

## Solution
The region is based on the sfincs.msk and since these were equal between the two versions I think we can safely replace the region.geojson in the gis folder.


DirkEilander:
I would suggest to also take this PR to modify the decimals of the geoms in `SfincsModel.write_vector` based on the crs. We do the same already in the `utils.write_xy`

While checking the code I also noticed that [in this line ](https://github.com/Deltares/hydromt_sfincs/blob/cab59fb3e6636c5d9482ecadcc85398ee60278c6/hydromt_sfincs/sfincs.py#L2817) `utils.write_xy` is called which does no longer exist. This should be `utils.write_xyn`.

roeldegoede:
> I would suggest to also take this PR to modify the decimals of the geoms in `SfincsModel.write_vector` based on the crs. We do the same already in the `utils.write_xyn`
> 
> While checking the code I also noticed that [in this line ](https://github.com/Deltares/hydromt_sfincs/blob/cab59fb3e6636c5d9482ecadcc85398ee60278c6/hydromt_sfincs/sfincs.py#L2817) `utils.write_xy` is called which does no longer exist. This should be `utils.write_xyn`.

Somehow the decimals seem to be introduced when writing the file .. I tried two methods for rounding before writing the gdf, but both methods resulted in lots of decimals in the geojson (even though the gdf looked rounded).

Method 1:
  # Convert to wkt and back to set precision
  wkt = gdf.geometry.to_wkt(rounding_precision=precision)
  gdf = gpd.GeoDataFrame(geometry=gpd.GeoSeries.from_wkt(wkt))

Method 2 (using https://gis.stackexchange.com/questions/321518/rounding-coordinates-to-five-decimals-in-geopandas):
gdf["geometry"] = gdf.geometry.apply(
    utils.round_coordinates, ndigits=precision
)


Coming back to your second point, utils.write_xy is not availbale because it is hosted in hydromt.io. I changed the reference to that functon, good catch!

DirkEilander (approved):

PR #143, opened 2023-11-08
Fix geographic crs

roeldegoede:
Some errors appeared for geographical coordinates, such as:

- `setup_waterlevel_bnd_from_mask` assumed the distance to be in meters, so too little bnd points were found for geographical coordinates.

- SFINCS needs the keyword crsgeo=1 for geographical coordinates

- Some errors in setup subgrid (to be investigated


DirkEilander:
Good improvement! In write_forcing (line 2982)  the format argument in the write_xy method should be updated based on the CRS.

PR #144, opened 2023-11-08
Add overviews to cloud optimized geotiffs

roeldegoede:


roeldegoede:
Do we want to have an argument for the overview levels?

DirkEilander (approved):
I've added two small suggestions, but generally it looks good to me.

hydromt_sfincs/utils.py
```diff
@@ -958,3 +958,33 @@ def _dist(a, b):
...
+    # check if fn is a geotiff file
+    extensions = [".tif", ".tiff"]
+    assert any(
+        fn.endswith(ext) for ext in extensions
+    ), f"File {fn} is not a GeoTIFF file."
+
+    # open rasterio dataset
+    with rasterio.open(fn, "r+") as src:
+        # create new overviews, resampling with average method
+        src.build_overviews([2, 4, 8, 16, 32], getattr(Resampling, resample_method))
```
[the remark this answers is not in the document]
roeldegoede: Great suggestion, added this to the code

hydromt_sfincs/utils.py
```diff
@@ -958,3 +959,55 @@ def _dist(a, b):
...
+    extensions = [".tif", ".tiff"]
+    assert any(
+        fn.endswith(ext) for ext in extensions
+    ), f"File {fn} is not a GeoTIFF file."
+
+    # open rasterio dataset
+    with rasterio.open(fn, "r+") as src:
+        # determine overviews when not provided
+        if overviews == "auto":
+            max_level = get_maximum_overview_level(src.width, src.height, 256)
```
DirkEilander, line 997: I would suggest to not hardcode the 256 but try to read it from the profile blockxsize argument.

```suggestion
            bs = src.profile.get('blockxsize', 256)
            max_level = get_maximum_overview_level(src.width, src.height, bs)
```

DirkEilander: please check if this code works before accepting, didn't test it!

roeldegoede: Works in my example, thanks for the suggestion!

Issue #127, opened 2023-09-13
Downscaling for large rotated models


roeldegoede:
For large rotated models, the auto-generated geotiff from the `setup_subgrid` routine cannot be read-in lazily. This is caused by the fact that for rotated grids, the coordinates also become 2D matrices instead of 1D vectors.

Potential solution (similar to how we dealt with memory issues in the subgrid routines):

    # Open the original geotiff
    input_geotiff_path = os.path.join(model,  "subgrid", "dep_subgrid.tif")

    block_size = 2000

    with rasterio.open(input_geotiff_path) as src:

        # Define block size
        n1, m1 = src.shape
        nrcb = block_size  # nr of cells in a block
        nrbn = int(np.ceil(n1 / nrcb))  # nr of blocks in n direction
        nrbm = int(np.ceil(m1 / nrcb))  # nr of blocks in m direction
        
        # avoid blocks with width or height of 1
        merge_last_col = False
        merge_last_row = False
        if m1 % nrcb == 1:
            nrbm -= 1
            merge_last_col = True
        if n1 % nrcb == 1:
            nrbn -= 1
            merge_last_row = True

        profile = dict(
            driver="GTiff",
            width=src.width,
            height=src.height,
            count=1,
            dtype=np.float32,
            crs=src.crs,
            transform=src.transform,
            tiled=True,
            blockxsize=256,
            blockysize=256,
            compress="deflate",
            predictor=2,
            profile="COG",
            nodata=np.nan,
            BIGTIFF="YES",  # Add the BIGTIFF option here
        )

        floodmap_tif = os.path.join(root, "floodmap.tif")
        with rasterio.open(floodmap_tif, "w", **profile):
            pass

        ## Loop through blocks
        ib = -1
        for ii in range(nrbm):
            bm0 = ii * nrcb  # Index of first m in block
            bm1 = min(bm0 + nrcb, m1)  # last m in block
            if merge_last_col and ii == (nrbm - 1):
                bm1 += 1

            for jj in range(nrbn):
                bn0 = jj * nrcb  # Index of first n in block
                bn1 = min(bn0 + nrcb, n1)  # last n in block
                if merge_last_row and jj == (nrbn - 1):
                    bn1 += 1

                # Define a window to read a block of data
                window = Window(bm0, bn0, bm1-bm0, bn1-bn0)

                # Read the block of data
                block_data = src.read(window=window)

                # check for nan-data
                if np.all(np.isnan(block_data)):
                    continue
[...]

PR #145, opened 2023-11-08
Fix downscaling large models

roeldegoede:
Note this PR builds upon #144  and it solves issue #127


hydromt_sfincs/utils.py
```diff
@@ -876,37 +881,133 @@ def downscale_floodmap(
...
+
+    elif isinstance(dep, (str, Path)):
+        assert (
+            floodmap_fn is not None
+        ), "floodmap_fn should be provided when dep is a Path or str."
+
+        with rasterio.open(dep) as src:
+            # Define block size
+            n1, m1 = src.shape
+            nrcb = 2000  # nr of cells in a block
```
DirkEilander, line 922: set as argument?

hydromt_sfincs/utils.py
```diff
@@ -876,37 +881,133 @@ def downscale_floodmap(
...
+            build_overviews(fn=floodmap_fn, resample_method="nearest")
+
+        hmax.name = "hmax"
+        hmax.attrs.update({"long_name": "Maximum flood depth", "units": "m"})
+        return hmax
+
+    elif isinstance(dep, (str, Path)):
+        assert (
+            floodmap_fn is not None
+        ), "floodmap_fn should be provided when dep is a Path or str."
```
DirkEilander, line 917: ValueError (or more specific errors) are considered better practice than assertions in python packages. arrestion errors are meant for testing / debuging.

```suggestion
        if floodmap_fn is not None:
            raise ValueError("floodmap_fn should be provided when dep is a Path or str.")
```

hydromt_sfincs/utils.py
```diff
@@ -876,37 +881,133 @@ def downscale_floodmap(
...
+                    # Define a window to read a block of data
+                    window = Window(bm0, bn0, bm1 - bm0, bn1 - bn0)
+
+                    # Read the block of data
+                    block_data = src.read(window=window)
+
+                    # check for nan-data
+                    if np.all(np.isnan(block_data)):
+                        continue
+                    # Convert row and column indices to pixel coordinates
```
DirkEilander, line 979: I don't think we need to do this now, but since we are working with rasterio (rather than xarray) it would make sense to also directly use the rasterio warp method rather than the raster.reproject method in `_downscale_floodmap_da`. I think this could provide quite a significant performance improvement. That would affect the whole block of code below this line.

roeldegoede: Interesting suggestion but let's postpone that for now. I added a TODO to the code

hydromt_sfincs/utils.py
roeldegoede: Good one, added an nrmax (similar as in setup_subgrid) argument.

PR #146, opened 2023-11-08
added Issue and PR templates

roeldegoede:
Added Issue and Pull-Request templates similar to HydroMT-core. This hopefully makes future issues a bit more clear.

Besides that, I added some pre-commit features to not worry anymore about black formatting, and I updated the zenodo information.


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,49 @@
+---
+name: Bug Report
+description: Report incorrect behavior in the HydroMT-SFINCS library
+labels: [Bug, Needs refinement]
```
DirkEilander, line 4: Not sure this is an issue but I don't think we have a "Needs refinement" label yet. To be sure we should make the issues we mention here at [this link](https://github.com/Deltares/hydromt_sfincs/labels). Idem for other templates

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+## Issue addressed
+Fixes #<issue number>
+
+## Explanation
+Explain how you addressed the bug/feature request, what choices you made and why.
+
+## Checklist
+- [ ] Updated tests or added new tests
+- [ ] Branch is up to date with `main`
+- [ ] Tests & pre-commit hooks pass
```
DirkEilander, line 10: we don't use pre-commit hooks yet in this repos.

.pre-commit-config.yaml
DirkEilander, line 1: I see you've added the pre-commit hooks settings. I suggest to do this a separate PR where we also apply all the styles. Or did you check it passes already? This would also require installing pre-commit (add to dev environment/ dependencies) and the hooks.

.zenodo.json
```diff
@@ -3,15 +3,16 @@
...
-            "affiliation":"Deltares",
-            "name": "de Goede, Roel"
-        },
         {
             "affiliation": "Deltares",
             "name": "Eilander, Dirk",
             "orcid": "0000-0002-0951-8418"
         },
+        {
+            "affiliation":"Deltares",
```
DirkEilander, line 12: happy to have you as first author on this software! Should we also add Maarten and Kees?

DirkEilander: Could you / did you also carefully compare this file against the hydromt core file to make sure it works?

DirkEilander (changes requested):
Nice improvement. I suggest to leave the pre-commit for another PR, see also comment below.  Besides adding new issue labels it looks good to me

.github/ISSUE_TEMPLATE/bug_report.yaml
roeldegoede: Done!

.github/pull_request_template.md
roeldegoede: Removed

.pre-commit-config.yaml
roeldegoede: Removed, let's discuss this later

.zenodo.json
roeldegoede: Wouldn't that be confusing, since it has been Eilander et al for a while?

File seems similar now to the core (though that already was the case), except the licensing.

DirkEilander: I think last time it failed on the license. It's hard to find good documentation about accepted license strings. I've changed based on another example I found which seemed to work.

DirkEilander (approved):

PR #147, opened 2023-11-09
bump hydromt core version

### LINKED ISSUE #142: Update to hydromt 0.9
Hello!

Would it be possible to update this package to be able to use hydromt 0.9?

Thank you!
Jens

DirkEilander:
## Issue addressed
Fixes #142 

## Explanation
Explain how you addressed the bug/feature request, what choices you made and why.

## Checklist


DirkEilander:
A silly typing issue now prevents compatibility with py39, see https://github.com/Deltares/hydromt/pull/655
Not sure if users are sill on that version though, could we already drop the support?

roeldegoede:
> A silly typing issue now prevents compatibility with py39, see [Deltares/hydromt#655](https://github.com/Deltares/hydromt/pull/655) Not sure if users are sill on that version though, could we already drop the support?

Think you can better judge this then me, but I'm not using 3.9 anymore, so for me dropping 3.9 supprt is fine. However, we can always bring it back right with a new update of hydromt-core when needed?

roeldegoede (changes requested):
Looks good! I think we want to bump the version number in the __init__.py to 1.0.3.dev and add this to the changelog as well?

Didn't add this myself since not sure if that is everything?

DirkEilander:
I've update the changelog and the version. We should can indeed add py3.9 back after the next core release, but as I don't think there are many py3.9 users I think it's ok to temporally drop support for now.

roeldegoede (approved):

PR #156, opened 2023-12-07
Change defaults

### LINKED ISSUE #154: mod.results['zsmax'] returns values much smaller than np.nanmax(mod.results['zs'])
### HydroMT-SFINCS version checks

- [X] I have checked that this issue has not already been reported.
- [X] I have checked that this bug exists on the latest version of HydroMT-SFINCS.

### Reproducible Example

If you want to reproduce it, download this [model run folder](https://drive.google.com/file/d/1mILeFjPxQG8qaPrSnxEyn8-vmqqsPEF7/view?usp=drive_link), and run the following code, replacing `sfincs_event_path` for the directory to the downloaded model folder:

```
mod = SfincsModel(sfincs_event_path, mode="r")
mod.read_results()

zsmax = mod.results['zsmax']
zsmax_alt = mod.results['zs'].max('time',skipna=True)

print('zs_max = ' + str(np.nanmax(zsmax.values)))
print('alternative zsmax = ' +str(np.nanmax(zsmax_alt.values)))
```

### Current behaviour

```
zs_max = 1.4580437
alternative zsmax = 7.2471085
```

### Desired behaviour

That the zsmax variable is similar or the same as taking zsmax from the max of the zs variables. 

I have produced a gif of the event and checked input BCs and the event appears to simulating as expected. There just seems to be a bug in the zsmax variable.

### Additional context

_No response_

roeldegoede:
## Issue addressed
- Fixes #154 for now. Default value is now equal to the SFINCS source code.
- Also set default for advection to 1 instead of 2 to avoid confusion, since advection=2 does the same as advection=1

## Explanation
If dtmaxout is larger then tstop-tstart, SFINCS writes the maximum outputs over the entire simulation. However, if dtmaxout is smaller than this, it only seems to write the outputs for the 2nd and higher intervals if the entire interval passed. The default value of the dtmaxout was 99999, which is only slightly more than 1 day. For simulations longer than 1 day, but less than x*99999, with the peak water levels in the x'th interval, this resulted in missing the peak water levels in the zsmax output.

## Checklist

The new default value might still be small? Are there more elegant ways to do this, instead of just making it even bigger? And shouldn't we ensure in SFINCS that even when the last interval is not fully finished, it still writes away the maximum outputs over the unfinished interval?


roeldegoede:
Are there more default values we should change?

roeldegoede:
To do: change dtmaxout in 1, 10 of 100 days instead of random big number that suggests to take the maximum over the entire simulation.

https://github.com/Deltares/SFINCS/pull/64 ensures that even for unfinished maximum-intervals at the end of the simulation, the maximum outputs are written in the file. Therefore, the maximum over the maximum output now also always is the maximum of the simulation.

So, I think dtmaxout=86400 (=1 day) would be a good new default value.

DirkEilander:
> So, I think dtmaxout=86400 (=1 day) would be a good new default value.

I've just updated this in when merging main into this branch.

DirkEilander (approved):

PR #160, opened 2023-12-15
Regular subgrid

### LINKED ISSUE #159: upgrade subgrid in SFINCS to v08
### Kind of request

### Enhancement Description

upgrade subgrid in SFINCS to v08

### Use case

_No response_

### Additional Context

_No response_

Leynse:
## Issue addressed
Fixes #159 

## Explanation
Copy paste code from quadtree_subgrid, without the quadtree part.
Bugfix leftover issues

## Checklist


roeldegoede:
The new NETCDF subgrid for regular grids still needs some attention. I think that this branch https://github.com/Deltares/SFINCS/tree/netcdf_subgrid_reggrid is a step into the right direction, but we have to check whether those changes are needed.

hydromt_sfincs/subgrid.py
```diff
@@ -18,13 +18,200 @@
...
+        var_list = ["zmin", "zmax", "ffit", "navg"]
+        for var in var_list:
+            uv_var = np.zeros(nr_points)
+            uv_var[index_mu1[active_indices]] = ds["u_" + var].values.flatten()[
+                active_cells
+            ]
+            uv_var[index_nu1[active_indices]] = ds["v_" + var].values.flatten()[
+                active_cells
+            ]
+            locals()["uv_" + var] = uv_var
```
DirkEilander, line 171: Why not storing it in `ds_new` directly? You would need to initialize `ds_new` earlier but I think that makes it more readable. I also suggest using f-strings to improve the readability e.g. 

`ds_new[f"uv_{var}"] = xr.DataArray(uv_var, dims=("npuv"))`

hydromt_sfincs/utils.py
```diff
@@ -1184,3 +1185,77 @@ def _downscale_floodmap_da(
         hmax = hmax.where(mask)
 
     return hmax
+
+
+def find_uv_indices(mask: xr.DataArray):
```
DirkEilander, line 1190: I'm not sure what indices are returned exactly. It would be good to add what the outputs represent and what types these are in the method.

hydromt_sfincs/subgrid.py
```diff
@@ -18,13 +18,200 @@
...
 
-    def load(self, file_name, mask):
+    # new way of reading netcdf subgrid tables
+    def read(self, file_name, mask):
+        """Load subgrid table from netcdf file."""
+
+        self.version = 1
+
+        # Read data from netcdf file with xarray
+        self.ds = xr.open_dataset(file_name)
```
DirkEilander, line 32: do we need to store ds to self, i.e. do we use it outside this method? 
I would keep it as a local variable and move the close the the end of the method.

hydromt_sfincs/sfincs.py
```diff
@@ -695,6 +697,8 @@ def setup_subgrid(
             by default 0
         nbins : int, optional
             Number of bins in which hypsometry is subdivided, by default 10
+        nlevels: int, optional
```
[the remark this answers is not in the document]
roeldegoede: Good one, done.

hydromt_sfincs/subgrid.py
roeldegoede: Agree, no need to store it indeed, so I changed it. Before, I was hoping that we would have something like SubgridTableRegular.data, where this data whas the xr.DataArray that could be visualized/used righ-away.

However, since SFINCS expects flattened arrays with only values for the active cells, and also combines the u and v points into the same arrays, the NetCDF file is still not as usefull as hoped. I added the reading-in of the subgrid tables yesterday, so now it works in the same way as the binary files... Think good for now, but would be nice when we could store the subgrid in a NetCDF that can be visualized immediately.

hydromt_sfincs/subgrid.py
roeldegoede: Good one, changed this!

hydromt_sfincs/utils.py
roeldegoede: Changed docstring into this, hopefully that clarifies your question?:

```
    """The subgrid tables for a regular SFINCS grid are organized as flattened arrays, meaning 
    2D arrays (y,x) are transformed into 1D arrays, only containing values for active cells.

    For the cell centers, this is straightforward, we just find the indices of the active cells.
    However, the u and v points are saved in combined arrays. Since u and v points are absent
    at the boundaries of the domain, the index arrays are used to determine the location of the
    u and v points in the combined flattened arrays.


    
    Parameters
    ----------
    mask: xr.DataArray
        Mask with integer values specifying the active cells of the SFINCS domain.

    Returns
    -------
    index_nm: np.ndarray
        Index array for the active cell centers.
    index_mu1: np.ndarray
        Index of upstream u-point in combined uv-array.
    index_nu1: np.ndarray
        Index of upstream v-point in combined uv-array.

    """
```

hydromt_sfincs/sfincs.py
```diff
@@ -2852,7 +2866,16 @@ def read_subgrid(self):
                 self.logger.warning(f"sbgfile not found at {fn}")
                 return
 
-            self.reggrid.subgrid.load(file_name=fn, mask=self.mask)
+            # re-initialize subgrid (different variables for old/new version)
+            # TODO: come up with a better way to handle this
+            self.reggrid.subgrid = SubgridTableRegular()
+            self.subgrid = xr.Dataset()
```
roeldegoede, line 2872: @DirkEilander you have suggestions here?

Since variables are slightly different between old/new methods, I think we always would like to start "clean" when we read a subgrid

tests/test_1model_class.py
```diff
@@ -103,6 +103,57 @@ def test_infiltration(mod):
...
+    # check version and new parameter
+    assert mod1.reggrid.subgrid.version == 0
+    assert "u_pwet" not in mod1.subgrid
+    assert "uv_pwet" not in mod1.subgrid
+    sbg_bin = mod1.subgrid.copy()
+
+    # compare z_zmin and z_zmax
+    assert np.isclose(np.sum(sbg_net["z_zmin"] - sbg_bin["z_zmin"]), 0.0)
+    # TODO: check with Maarten whether this is meant to be different
+    # difference comes from different discretization of volume bins
```
roeldegoede, line 153: @DirkEilander this is still something I would like to discuss with Maarten

hydromt_sfincs/utils.py
DirkEilander: Thanks! This give more context to read the code. I asked because I think that it might be easier (and faster as it saves many python loops) to get the 2D indices fashion by stacking the mask and shifted masks for u and v to find all valid indices. But I don't have sufficiently thorough understanding of the method to refactor it.

hydromt_sfincs/sfincs.py
DirkEilander: I think this is ok. But maybe I don't understand your concern. We can chat about it next week?

roeldegoede: Okay, then it's good, since it does what I was aiming for. I was just hoping that there would be more elegant ways to "re-initialize" these objects in the self.reggrid.subgrid.read_(binary) methods.

hydromt_sfincs/utils.py
roeldegoede: Yes, agree on this, but I couldn't further optimize this.

Think the main problem here is that there are no u- and/or v point at the outer edges of the domain, which results sometimes in two u-points in a row and sometimes two v-points in a row, whereas the usually are alternating (in the combined uv matrix)

PR #161, opened 2023-12-20
fix double reading

DirkEilander:
## Issue addressed
Fixes reading some files double, see also https://github.com/Deltares/hydromt/issues/694

## Explanation
While this is also fixed in the [core](https://github.com/Deltares/hydromt/pull/695/files), this fix will temporarily achieve the same here. We can replace the changes here with the initialize_xxx methods introduced in core..

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,7 @@ Changed
 
 Fixed
 -----
+- fix a bug that caused some model files to be read twice (#102)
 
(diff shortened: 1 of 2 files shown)


roeldegoede (approved):
Nice catch and good solution! Only changed the PR you mentioned in the changelog, since you were mentioning 102 and it should be 161?

PR #162, opened 2023-12-20
fix filter timeseries in setup_x_forcing

DirkEilander:
## Issue addressed
Fixes filtering of timeseries in case you have a large dataset in separate files for the timeseries and locations (rather than one geodataset). If the locations are filtered based on the model region a subset of the locations and all timeseries are passed to the `set_1d_forcing` method raising an error that the indices do not match.

## Explanation
Add a filtering step after reading the location in `setup_waterlevel_forcing` and `setup_discharge_forcing`

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,7 @@ Added
 Changed
 -------
 - support for HydroMT core v0.9
+- support for filtering forcing data from timeseries and locations based on model region (#103)
 
 Fixed
 -----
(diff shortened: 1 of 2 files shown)


docs/changelog.rst
```diff
@@ -15,6 +15,7 @@ Added
 Changed
 -------
 - support for HydroMT core v0.9
+- support for filtering forcing data from timeseries and locations based on model region (#103)
```
roeldegoede, line 18: Am I looking wrong or should this also be #162 instead of #103?

roeldegoede (changes requested):

roeldegoede (approved):

PR #183, opened 2024-02-13
Docstring for setup_grid_from_region bbox option needs to be in WG84

### LINKED ISSUE #182: Setup grid from region with bounding box in different CRS
### HydroMT-SFINCS version checks

- [X] I have checked that this issue has not already been reported.
- [X] I have checked that this bug exists on the latest version of HydroMT-SFINCS.

### Reproducible Example

I am trying to setup the grid with a bounding box that is already in a different CRS e.g. like this

```python
bbox = [406565.1340,4567290.4902,436760.4489,4585912.0943]

sf.setup_grid_from_region(
    dict(bbox=bbox),
    res=200,
    crs=32631,
)
```
Which I think leads to the self.geoms being full of `inf`

### Current behaviour

It fails with `ValueError: cannot convert float NaN to integer` in *hydromt_sfincs/sfincs.py:303*

### Desired behaviour

I think in L408 of model_api.py where the bounding box is created via GeoPandas it could probably use a dynamic CRS?

### Additional context

_No response_

saschahofmann:
Closes #182 

I just added a short note.
> For the ``bbox`` option the coordinates for the bottom-left and top-right corners
>           need to be provided in WG84/EPSG:4326.

I also noticed that there seems to be a formatting error https://deltares.github.io/hydromt_sfincs/latest/_generated/hydromt_sfincs.SfincsModel.setup_grid_from_region.html
```
For a complete overview of all region options, see :py:function:~hydromt.workflows.basin_mask.parse_region
```
but I am not really familiar with the Sphinx syntax, not sure whether it is supposed to link somewhere or whether `~` should have been `:` which is what I went for.

### DIFF
--- a/hydromt_sfincs/sfincs.py
+++ b/hydromt_sfincs/sfincs.py
@@ -239,8 +239,11 @@ class SfincsModel(GridModel):
             * {'bbox': [xmin, ymin, xmax, ymax]}
             * {'geom': 'path/to/polygon_geometry'}
 
+            For the ``bbox`` option the coordinates for the bottom-left and top-right corners
+            need to be provided in WG84/EPSG:4326.
+            
             For a complete overview of all region options,
-            see :py:function:~hydromt.workflows.basin_mask.parse_region
+            see :py:function:: hydromt.workflows.basin_mask.parse_region
         res : float, optional
             grid resolution, by default 100 m
         crs : Union[str, int], optional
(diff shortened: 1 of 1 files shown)


roeldegoede:
Thanks for your work! I think I now fixed the reference to the hydromt-core function, but there are still some tests failing.

The failing of these tests is most likely unrelated to these changes, but I will have a look at this next week.

roeldegoede (approved):

roeldegoede:
tests will be fixed in a different branch :)

PR #200, opened 2024-06-05
update weigths between zmin and zmax in uv points

roeldegoede:


roeldegoede:
We still need a good description of what has been changed from "old" to "new", and is max_gradient = 999 a logical value?

Leynse (approved):

PR #202, opened 2024-06-18
move inflow points to confluence if within buffer from model boundary

DirkEilander:
## Issue addressed
Fixes issue with multiple inflow points near confluence. 
In case of a confluence near the model region I want to be able to select the confluence (see figure b) rather than both tributaries as inflow points (see figure a). This is achieved by removing small river segments that completely fall within a user-defined buffer from the model boundary. The most upstream point of the remaining river is than located at the confluence.

**figure a: inflow at to closeby tributaries before fix**
image

**figure b: inflow at confluence after fix**
image

## Checklist

### DIFF
--- a/hydromt_sfincs/workflows/flwdir.py
+++ b/hydromt_sfincs/workflows/flwdir.py
@@ -82,7 +82,7 @@ def river_source_points(
     gdf_riv: gpd.GeoDataFrame,
     gdf_mask: gpd.GeoDataFrame,
     src_type: str = "inflow",
-    buffer: float = 100,
+    buffer: float = 200,
     river_upa: float = 10,
     river_len: float = 1e3,
     da_uparea: xr.DataArray = None,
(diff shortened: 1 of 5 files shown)


hydromt_sfincs/workflows/flwdir.py
```diff
@@ -143,6 +144,11 @@ def river_source_points(
 
     # clip river to model gdf_mask
     gdf_riv = gdf_riv.to_crs(gdf_mask.crs).clip(gdf_mask.unary_union)
+    # keep only lines
+    gdf_riv = gdf_riv[
+        [t.endswith("LineString") for t in gdf_riv.geom_type]
+    ].reset_index(drop=True)
```
roeldegoede, line 150: Due to this reset_index, the tests were failing. I now fixed it on the test side, but we could also consider fixing it here. Points returned by the code were identical, only the riv_idx weren't.

DirkEilander: decided to remove riv_idx and test on the coordinates instead.

roeldegoede (approved):

PR #208, opened 2024-07-31
Default for baro should be set to 1! As is default in SFINCS itself, …

Leynse:
Default for baro should be set to 1! As is default in SFINCS itself, otherwise any pressure input is ignored....


roeldegoede:
Should we add/modify/remove any other defaults?

PR #210, opened 2024-08-28
Parse forcing precision

### LINKED ISSUE #197: write forcing precision
### Kind of request

### Enhancement Description

Hi, I just wanted to make a suggestion for a modification to the write_forcing() function in sfincs.py. When saving the hydrograph data in the .bzs file I miss the possibility to specify the precision with which to save the data.

Although it is true that the write_timeseries() function of utils.py can be called separately to specify the precision, it would be nice if the write_forcing() function could handle this option.

### Use case

For example, it could be an attribute of the class, so that when calling the SfincsModel.write() function, if the precision for the .bzs file has been previously specified, the file would be saved with the desired precision without the need to add arguments to the functions.

### Additional Context

_No response_

roeldegoede:
## Issue addressed
Fixes #197 

## Explanation
By default still two decimals are used, but you can overwrite the fmt in SfincsModel.write_forcing.

Another option is to provide fmt="auto", then the format is derived from the forcings.

## Checklist
- [x ] Updated tests or added new tests


DirkEilander (commented):
Looks good to me! Note that it does not completely solve #197. We should also respond in the issue what we will do with the idea by to store the forcing format as an attribute of SfincsModel. I think this has low priority for us, but we can welcome a contribution on this from the issue owner.

DirkEilander (approved):

PR #213, opened 2024-08-29
Prepare v1.1.0

roeldegoede:
Preparation of new release

Updated changelog
Solved most deprecation warnings
Updated dependencies

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,8 +12,8 @@ authors = [
 ]
 dependencies = [
     "affine",
-    "geopandas>=0.8",
-    "hydromt>=0.9.1, <0.11",
+    "geopandas>=1.0",
+    "hydromt>=0.10, <0.11",
     "numba",
     "numpy<2.0", # temp pin until bottleneck release v1.4
     "pandas",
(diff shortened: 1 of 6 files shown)


roeldegoede:
Only open point is changelog (adding date and info of the release), then we can make a release.

pyproject.toml
```diff
@@ -12,8 +12,8 @@ authors = [
 ]
 dependencies = [
     "affine",
-    "geopandas>=0.8",
-    "hydromt>=0.9.1, <0.11",
+    "geopandas>=1.0",
+    "hydromt>=0.10, <0.11",
```
DirkEilander, line 16: I think you can leave the lower version limit unchanged right? It will by default take the latest, but I think everything should work with 0.9 as well. 
```suggestion
    "hydromt>=0.9, <0.11",
```

roeldegoede: Seems like we have quite some nice fixes in 0.10.0? MOstly related to bugs in raster.py for rotated grids, working with zoomlevels etc

https://github.com/Deltares/hydromt/releases

Leynse (approved):

PR #220, opened 2024-10-21
clean grid variables when recreating a grid

roeldegoede:
## Issue addressed
When re-running the setup_grid (or setup_grid_from_region) functions after you added a gridded varaible, dep or msk for example, the RegularGrid class was updated correctly, but the SfincsModel.grid seemed to stick on the first grid and therefore the model didn't update correctly.

Not sure when this was introduced, but espeically in a GUI, refining the grid after adding variables happens often, so this is a serious bug ...

## Explanation
setup_grid is always called when creating a new grid, so forcing the SfincsModel.grid to start blanco should solve the problem

## Checklist

### DIFF
--- a/envs/hydromt-sfincs-dev.yml
+++ b/envs/hydromt-sfincs-dev.yml
@@ -8,7 +8,7 @@ dependencies:
 - cartopy
 - descartes
 - ffmpeg
-- geopandas>=1.0
+- geopandas>1.0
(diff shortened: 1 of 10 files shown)


roeldegoede:
Only tests that are breaking now are with the latest HydroMT-core version, but that is supposed to break I guess. So I think this can be merged

PR #224, opened 2024-11-27
Downscale floodmap

vanasseltk:
## Explanation
Small bug fixes in downscale floodmap function.


roeldegoede (approved):
Looks good. Tests will be fixed when #220 is merged

PR #227, opened 2024-12-06
Added factor for ks

keesnederhoff:
## Issue addressed
No issue, but extra functionality

## Explanation
Added a factor to covert ks - (e.g. from micrometer per second to mm/hr which is SFINCS required)

## Checklist
- [] Updated tests or added new tests [not applicable]
- [] Branch is up to date with `main` [not applicable]
- [] Updated documentation if needed [not applicable]
- [ ] Updated changelog.rst if needed  [not sure where that is]


hydromt_sfincs/sfincs.py
```diff
@@ -1342,6 +1342,9 @@ def setup_cn_infiltration_with_ks(
 
         # Done
         self.logger.info(f"Done with determination of values (in blocks).")
+        
+        # Convert ks - (e.g. from micrometer per second to mm/hr which is SFINCS required)
+        da_ks = da_ks * factor_ksat
```
roeldegoede, line 1347: Looks fine to me, though this is also what the data_catalog is meant for (changing variable names to the expected ones, changing units etc), for example here for era5 data with unit_mult:

```
era5_hourly:
  crs: 4326
  data_type: RasterDataset
  driver: netcdf
  kwargs:
    chunks:
      latitude: 125
      longitude: 120
      time: 50
    combine: by_coords
    decode_times: true
    parallel: true
  meta:
    category: meteo
    notes: Extracted from Copernicus Climate Data Store
    paper_doi: 10.1002/qj.3803
    paper_ref: Hersbach et al. (2019)
    source_license: https://cds.climate.copernicus.eu/cdsapp/#!/terms/licence-to-use-copernicus-products
    source_url: https://doi.org/10.24381/cds.bd0915c6
    source_version: ERA5 hourly data on pressure levels
  path: meteo/era5/{variable}/era5_{variable}_{year}_hourly.nc
  rename:
    msl: press_msl
    ssrd: kin
    t2m: temp
    tisr: kout
    tp: precip
  unit_add:
    temp: -273.15
  unit_mult:
    kin: 0.000277778
    kout: 0.000277778
    precip: 1000
    press_msl: 0.01
```

hydromt_sfincs/sfincs.py
```diff
@@ -2803,7 +2806,7 @@ def write(self):
         # config last; might be udpated when writing maps, states or forcing
         self.write_config()
         # write data catalog with used data sources
-        self.write_data_catalog()  # new in hydromt v0.4.4
+        #self.write_data_catalog()  # new in hydromt v0.4.4
```
roeldegoede, line 2809: Why was this needed?

hydromt_sfincs/sfincs.py
roeldegoede: Nonetheless, fine with this extra functionality, just for your info.

keesnederhoff: Thanks for sharing this information, very helpful. However, for users that don't utilize the data catalog still a useful functionality, so let's add this anyway.

hydromt_sfincs/sfincs.py
keesnederhoff: Cannot completely remember anymore, but the writing of the data catalog often results in issues, so my suggestions is to not have this on by default

roeldegoede: I believe quite some users like this functionality since it highly improves reproducability (even if you don't know anything about data catalogs). I do recall that it sometimes failed, especially with datasets that have a less commonly used CRS. However, it's the last function called in "write", so the rest of the model is already written and works well.

I like the option of making it optional, with a keyword "write_data_catalog", but I would put the default to True.

roeldegoede:
Changed optional argument into try and except. Considered to also remove the data_catalog file that is written, but since it is not automatically read-in (and therefore not causing issues later on), I kept it.

roeldegoede (approved):
Wait with merging untill the other PRs are merged to see if any tests are broken

PR #241, opened 2025-01-28
Overwriting files that were lazily loaded results in permission error…

roeldegoede:
…s. Loading the data first solves that problem.

## Issue addressed
When reading in a sfincs model with netcdf files (as grid (future)), subgrid or as forcing), these datasets are lazily loaded. In theory, that increases performance, because data is only loaded when necesssary. However, due to lazy references to the data in the original files, the files cannot be overwritten when you write the model.

Of course, you could choose to only write the files that you changed, but often the SfincsModel.write() is used.

## Explanation
Everywhere where we would like to write a netcdf, we first check if the file already exists. If that is the case, we have to load the data into memory first, such that we can overwrite the file after (after loading, the lazy references are gone).

## Checklist
- [x ] Updated tests or added new tests
- [x ] Branch is up to date with `main`

Question is whether we even want to load and overwrite datasets that are still lazy, because in most cases that will mean that they are unchanged and loading/writing would result in unnecessary computations?


roeldegoede:
Somehow, the tests that I implemented to guarantee this would work (which it perfectly did locally) fail online with he same permission errors as before. However, since everything works on my side, it's very hard to debug..  @DirkEilander or @LuukBlom any suggestions are welcome.

tests/test_quadtree.py
```diff
@@ -35,3 +38,34 @@ def test_quadtree_io(tmpdir):
...
+
+    # Convert to dataset
+    ds = ds.ugrid.to_dataset()
+
+    # Try to write
+    # NOTE this should fail because it still has lazy references to the file
+    try:
+        ds.to_netcdf(nc_copy)
+    except PermissionError:
+        pass
```
LuukBlom, line 62: `to_netcdf()` can succeed or fail, doesn't affect the test at all?

use 
```python
with pytest.raises(PermissionError) as e:
  ds.to_netcdf(nc_copy)
  assert .... in str(e)
```

tests/test_quadtree.py
```diff
@@ -35,3 +38,34 @@ def test_quadtree_io(tmpdir):
...
+
+    # Create file + copy
+    shutil.copy(ncfile, nc_copy)
+
+    # Open the copy with xu_open_dataset
+    # This opens the file lazily
+    ds = utils.xu_open_dataset(nc_copy)
+
+    # Convert to dataset
+    ds = ds.ugrid.to_dataset()
```
LuukBlom, line 55: here you are overwriting the `ds` variable with another object, which could trigger the garbage collector and release the files. m not sure if this breaks anything, but I would stay away from using the same variable name for two different objects.

hydromt_sfincs/utils.py
```diff
@@ -1338,3 +1341,29 @@ def xu_open_dataset(*args, **kwargs):
...
+
+    # Check for lazy loading
+    lazy_vars = [not data_array._in_memory for data_array in ds.data_vars.values()]
+
+    # if all(lazy_vars):
+    #     return  # All variables are lazy-loaded, skip writing?
+
+    if any(lazy_vars):
+        ds.load()  # Some variables are lazy-loaded, load them into memory
+    return
```
LuukBlom, line 1369: Make sure to call `ds.close()` the file after loading the contents

tests/test_quadtree.py
roeldegoede: Good one, I was looking for something like that, but then forgot

tests/test_quadtree.py
roeldegoede: Pretty sure this happens in more places in the code, I could change that.

hydromt_sfincs/utils.py
roeldegoede: We discussed earlier that ds.close() does not work for xugrid.UgridDatasets right?

LuukBlom: Ah yes, I keep forgetting. 
Unfortunately, for the same reason `ds.load()` will also not work then :s

PR #247, opened 2025-03-25
fix missing projection method

jensdebruijn:

PR #225, opened 2024-11-27
Bugfix havg

roeldegoede:
## Issue addressed
Bugfix in new subgrid implementation

## Explanation
Explain how you addressed the bug/feature request, what choices you made and why.

## Checklist
- [ ] Updated tests or added new tests
- [ ] Branch is up to date with `main`
- [ ] Updated documentation if needed
- [ ] Updated changelog.rst if needed

## Additional Notes (optional)
Add any additional notes or information that may be helpful.

PR #250, opened 2025-04-02
first_commit

### LINKED ISSUE #249: River Bathy using gdf_zb
### Kind of request

None

### Enhancement Description

I am setting up a inland river model. I am using the burn_in_rect function and geodataframe with riverbed points. If you supply a gdf_zb, this file is not clipped to the subgrid block size. This results in infinite values in the interpolation. 

### Use case

_No response_

### Additional Context

_No response_

vanasseltk:
## Issue addressed
Fixes #249

## Explanation
Clipping gdf_zb such that only points that are actually in the subgrid block are included.

## Checklist

When there is no points in the block, you do not have any points to use for interpolation. This will also result in a nan value right?


vanasseltk:
@roeldegoede nieuwe commit gedaan, die tests blijven falen, wat kan ik daar aan doen?

hydromt_sfincs/workflows/bathymetry.py
```diff
@@ -307,13 +311,16 @@ def burn_river_rect(
...
         # TODO manning always defined on centerline?
+
         if manning_name not in gdf_zb.columns and manning_name in gdf_riv.columns:
             gdf_zb[manning_name] = np.nan
-        if np.any(np.isnan(gdf_zb[manning_name])):
-            gdf_zb["idx0"], _ = nearest(gdf_zb, gdf_riv)
-            man_nearest = gdf_riv.loc[gdf_zb["idx0"], manning_name]
-            man_nearest.index = gdf_zb.index
-            gdf_zb[manning_name] = gdf_zb[manning_name].fillna(man_nearest)
+        if manning_name in gdf_zb.columns:
```
roeldegoede, line 317: is this 2nd if manning_name in gdf_zb.columns needed? I think the code that follows can (and should) only be executed when gdf_zb[manning_name] has nans, **and** gdf_riv[manning_name] exists .. if the latter is not the case, the code crashes no?

vanasseltk: Yes, i agree, but i got an error here because the key "manning_name" was not present in gdf_zb. So that can de expained if manning_name in gdf_riv.columns is not present, and that is possible right? Previously it was only if np.any(np.isnan(gdf_zb[manning_name])), at the same level as  if np.any(np.isnan(gdf_zb[manning_name])). 

Long story short, an indent without the if statement will do.

roeldegoede (approved):
Looks good. Tests are failing because of #247. Those are already fixed as part of #225.

PR #263, opened 2025-06-13
add creation of index cog and optimzie downscale_floodmap accordingly

roeldegoede:
The creation of indices could speed-up the process of downscaling. First tests showed  speedups for unstructured grids, especially for in-memory downscaling. First comparisons for structured (non rotated) grids didn't show big differences.


Leynse (approved):
- Worked well for my real life case in Denmark, nice reduction in computation time!

PR #287, opened 2025-10-08
V1/elevation

roeldegoede:
In this PR we renamed datasets_dep into elevation_sets and in accordance, we named datasests_rgh and datasets_riv, roughness_sets and river sets respectively. All tests and examples are updated.

In additions, I added back the bathymetry database for quadtree creation and elevation, not tested properly yet ... For testing and comparing it can be nice to keep both methods of bringing in the data, but in the long run, might be better to work towards one data provided. I think #282 provides the possibility of working with TiledWebMaps in HydroMT as well.

For now:
- _quadtree_grid.create_ has the option of providing a bathymetry database 
- _quadtree_grid.elevation.create_ works only for HydroMT data and uses xugrid partitioning and interpolation -- _quadtree_grid.elevation.set_bathymetry_ should work with bathymetry_database and HydroMT. This uses interpolation methods for regular grids and might be faster. In the long run, we want either one of the two functions named "create", but for now easy to test around a bit.

### DIFF
--- a/examples/build_quadtree_from_script.ipynb
+++ b/examples/build_quadtree_from_script.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## EXAMPLE Quadtree"
+    "## Example Quadtree"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


roeldegoede (commented):
elevation_sets --> elevation_list?
and similar?

PR #294, opened 2025-10-30
Enhancements/pringle01

WPringle:
## Explanation
Suggested changes that I made through the development of my use case with subgrid features:
- topobathy is the U.S. CUDEM
- land use data (to convert to Manning's) is something something.

The data is complete without NaNs except beyond a certain depth/distance from coastline as shown in plot below. I use this point (where becomes NaN in deep water) to define the boundary of the SFINCS model. So there is no need to e.g., merge with GEBCO bathy data or check for missing values. This feature of the data and model is a key reason for the needed changes in this PR.

Additionally, a couple of other miscellaneous changes:
1. Force `compute.()` of `xarray`'s lazy implementation of the conversion of land use table to Manning's n when generating the subgrid information. This is for speed. Without this addition, the same computation will be repeated many times over when using a large dataset.
2.  In `plot_forcing` for plotting wind, I changed the string catch simply to `startswith(wind)` which will catch all wind forcing options (otherwise misses `wind10_u` and `wind10_v` forcing).

## Checklist

### DIFF
--- a/hydromt_sfincs/regulargrid.py
+++ b/hydromt_sfincs/regulargrid.py
@@ -189,6 +189,8 @@ class RegularGrid:
         da_mask0 = None
         if not reset_mask and da_mask is not None:
             # use current active mask
+            da_mask2 = da_mask == 2  # for keeping these values
+            da_mask3 = da_mask == 3  # for keeping these values
             da_mask0 = da_mask > 0
         elif gdf_mask is not None:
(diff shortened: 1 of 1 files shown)


WPringle:
@roeldegoede Thanks for the information! Sure, I will work on explaining the PR a bit and finalizing it.

PR #318, opened 2026-01-14
quick and dirty fix for plotting boundaries for quadtree models .. mi…

### LINKED ISSUE #239: Improve the way boundaries are plotted in plot_basemap
### Kind of request

### Enhancement Description

We now first vectorize the msk=2 or msk=3 cells into a line, before we plot the boundaries in the plot basemap. This doesn't work for unstructred grids (future work), and seems a bit too complex in general for very simple functionality.

- We could convert the msk=2 cells to "dots", so no conversion to linestrings is needed
- We could consider plotting the raster/unstructred mask, but only the boundary cells, replacing the normal active domain (msk=1) with nans.

### Use case

_No response_

### Additional Context

_No response_

roeldegoede:
…ght get slow for big models

## Issue addressed
Fixes #239 

Not the nicest way of solving this I guess, but it seems to work for now ..

### DIFF
--- a/hydromt_sfincs/plots.py
+++ b/hydromt_sfincs/plots.py
@@ -4,7 +4,6 @@ import logging
 from typing import Dict, List, Tuple, Union
 
 import numpy as np
-from matplotlib.lines import Line2D
 import pandas as pd
 import xarray as xr
 import xugrid as xu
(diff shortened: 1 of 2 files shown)


roeldegoede:
@Leynse what do you think? We could use this utils function now also for waterlevel and snapwave boundaries! It creates linestrings, so the only thing left is sampling points along the lines!

PR #329, opened 2026-01-26
Fix/wave bc

Leynse:
## Issue addressed
Fixes #<issue number>

## Explanation
Explain how you addressed the bug/feature request, what choices you made and why.

## Checklist

### DIFF
--- a/hydromt_sfincs/components/forcing/boundary_conditions.py
+++ b/hydromt_sfincs/components/forcing/boundary_conditions.py
@@ -121,6 +121,10 @@ class SfincsBoundaryBase(ModelComponent):
             if geodataset.vector.crs != self.model.crs:
                 geodataset = geodataset.vector.to_crs(self.model.crs)
             # keep dataset dims ordering consistent
+            index_dim = geodataset.vector.index_dim
+            if index_dim != "index":
+                # rename the index dimension to "index" if needed
+                geodataset = geodataset.rename({index_dim: "index"})
             self._data = geodataset.transpose("time", "index", ...)
             return
 
@@ -406,7 +410,7 @@ class SfincsBoundaryBase(ModelComponent):
                     dims=("time", "index"),
                     coords={
                         "time": time,
-                        "index": gdf.index,
+                        "index": gdf.index.values,
                     },
                     name=var,
                 )
--- a/hydromt_sfincs/components/forcing/water_level.py
+++ b/hydromt_sfincs/components/forcing/water_level.py
@@ -446,6 +446,8 @@ class SfincsWaterLevel(SfincsBoundaryBase):
             )
             df_ts = da.transpose(..., da.vector.index_dim).to_pandas()
             gdf_locs = da.vector.to_gdf()
+            if gdf_locs.index.name == "stations":
+                gdf_locs.index.name = "index"
         elif timeseries is not None:
             df_ts = self.data_catalog.get_dataframe(
                 timeseries,


hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -83,184 +105,96 @@ def read_boundary_points(self, filename: str | Path = None):
...
-        # # Add to self.data
-        # self.data = gdf
-
-        # Read the bnd file
-        df = pd.read_csv(
-            abs_file_path, index_col=False, header=None, names=["x", "y"], sep="\s+"
-        )
+        # TODO check if we want read_xyn? Before we used read_xy, so without name column
+        gdf = utils.read_xyn(abs_file_path, crs=self.model.crs)
+        gdf = gdf.set_crs(self.model.crs)
```
roeldegoede: crs already set in utils.read_xyn

hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -273,22 +207,42 @@ def write(self, format: str = None):
...
+                "snapwave_bdsfile",
+            ]
+            vars = [
+                "hs",
+                "tp",
+                "wd",
+                "ds",
+            ]  # FIXME - should be 'hs' or 'bhs'? in kernel is 'hs', but it's 'bhsfile'...
+            for i, varname in enumerate(filenames):
+                self.write_boundary_conditions_timeseries(var=vars[i], varname=varname)
```
roeldegoede: A bit confusing that we use filename for varname (is that correct?), while the function also has filename as an argument? Also, with the vars as a list, you could also easily construct the filname in the loop, that would save some lines of code

```
for var in ["hs", "tp", "wd", "ds"]:
    filename = f"snapwave_{var}file"
    self.write_boundary_conditions_timeseries(var=var, varname=filename)
```

hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -307,130 +261,363 @@ def write_boundary_points(self, filename: str | Path = None):
...
+        set
+        """
+        gdf_locs, df_ts = None, None
+        tstart, tstop = self.model.get_model_time()  # model time
+        vars = ["hs", "tp", "wd", "ds"]
+        # buffer around msk==2 values
+        if not self.model.grid_type == "quadtree":
+            raise ValueError("SnapWave is not supported for regular grid models!")
+        else:
+            region = self.model.region
```
roeldegoede: With the latest additions of #318, we should be able to retrieve the snapwave_mask=2 values and buffer around it, if we like that feature ...

hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -307,130 +261,363 @@ def write_boundary_points(self, filename: str | Path = None):
...
+
+        # read wave data from geodataset or geodataframe
+        if geodataset is not None:
+            # read and clip data in time & space
+            da = self.data_catalog.get_geodataset(
```
roeldegoede: for consistency better to name a dataset ds instead of da?

hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -307,130 +261,363 @@ def write_boundary_points(self, filename: str | Path = None):
...
+                        # df_ts.columns.name = "index"
+                        df_ts.columns = df_ts.columns.map(
+                            int
+                        )  # parse column names to integers
+
+                        # set per variable
+                        self.set_timeseries(
+                            df=df_ts,
+                            varname=vars[i],
+                        )
```
roeldegoede: I guess this is tricky ... Seems that all timeseries are replaced when "len(new_da.indexes["index"]) == self.nr_points". So that would mean that each variable overwrites the previous one. I guess we should update the set_timeseries to immediately accept an xarray?

hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -52,25 +50,49 @@ def read(self, format: str = None):
...
+                        var=vars[i], varname=varname
+                    )
+
+                    da = xr.DataArray(df, dims=("time", "index"), name=vars[i])
+                    da_lst.append(da)
+
+                # assumed is that all timestamps and number of points are the same in all files
+                # (as they should be for SFINCS-SnapWave kernel)
+                ds = xr.merge(da_lst[:])
+
```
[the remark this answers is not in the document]
Leynse: Moved this code to new function 'read_boundary_conditions_timeseries'

hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -83,184 +105,96 @@ def read_boundary_points(self, filename: str | Path = None):
...
+                raise ValueError(
+                    f"Variable {var} not found in SnapWave boundary conditions netcdf file!"
+                )
 
-        ds.close()
+        # Rename variables to match internal naming
+        ds = ds.rename({"stations": "index"}) if "stations" in ds.dims else ds
+        return ds
+
+        # # Loop through boundary points
```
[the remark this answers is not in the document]
Leynse: No is still needed, because in boundary_conditions 'set()'. the dimension 'index' is expected, while from the data read from the netcdf bc file it is still 'stations'

hydromt_sfincs/components/forcing/snapwave_boundary_conditions.py
```diff
@@ -591,179 +713,180 @@ def set_timeseries(
...
-                break
+        # Call set_timeseries to update your object's data
+        self.set_timeseries(df_hs, varname="hs")
+        self.set_timeseries(df_tp, varname="tp")
+        self.set_timeseries(df_wd, varname="wd")
+        self.set_timeseries(df_ds, varname="ds")
 
-            # Find first the unused points
-            i1 = np.where(~used)[0][0]
+    def create_boundary_points_from_mask(self, min_dist=None, bnd_dist=5000.0):
```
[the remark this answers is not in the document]
Leynse: Added to issue longlist

PR #352, opened 2026-03-11
some minor quadtree enhancements for missing and interpolating data

roeldegoede:
Minor bugfixes that appeared when creating quadtree models with datasets that didn't cover the full domain.


vanasseltk (approved):
Tested for two quadtree model. all looks good. 

Just wondering how we never notice the double indexing wehne reding in quadtree.

PR #356, opened 2026-03-24
340 clean up create boundary points from mask functions

### LINKED ISSUE #340: Clean up create_boundary_points_from_mask functions
### Kind of request

### Enhancement Description

- For both water level and wave.
- Move overlapping parts to separate function

### Use case

_No response_

### Additional Context

_No response_

roeldegoede:
Still to discuss:

- In the old HydroMT-SFINCS code, we always used absolute distances, so with a length of 5k, and a boundary distance of 2km, we would get bnd points at 0,2 and 4km. 
- In the cht_sfincs code, the the start and end of the boundary line get a point, and in between you approximate the boundary distance. So you would get bnd poitns at 0,1.67,3,33 and 5km probably. 

Now I included both, but that seems a bit like an overkill. Do we have a preference?


tests/test_water_level_boundary.py
```diff
@@ -298,6 +298,18 @@ def test_create(model_config):
     assert model_config.water_level.data["bzs"].index[-1] == 0
 
 
+@pytest.mark.parametrize("model_fixture", ["model", "quadtree_model"])
+def test_create_boundary_points_from_mask(request, model_fixture):
```
Leynse, line 302: Overkill to add this as well to test_snapwave_boundary_conditions?

hydromt_sfincs/utils.py
```diff
@@ -471,6 +478,51 @@ def get_bounds_vector(
...
+    Returns
+    -------
+    GeoDataFrame
+        Points along the input line geometries.
+    """
+    points = []
+
+    for _, row in gdf_lines.iterrows():
+        line = row.geometry
+        # TODO discuss whether these methods are needed, probaly we could just pick one
```
Leynse: I see your point of both methods.

I could imagine doing 1 hybrid of both: do "absolute", but always also add a point at the end (even if that leaves e.g. a 1km section)

Leynse (changes requested):
- See my 2 remarks, for the rest I approve - nice and clean addition!

tests/test_water_level_boundary.py
roeldegoede: Guess so? It's almost identical functionality now

Leynse (approved):
- Great

PR #358, opened 2026-03-25
add back discharge_forcing_from_grid

### LINKED ISSUE #327: Add back setup_river_discharge_from_grid()
### Kind of request

Adding new functionality

### Enhancement Description

In HydroMT-SFINCs versions <2.0.0, we had a function called setup_discharge_forcing_from_grid. We didn't copy that in the newer versions, and it turned out to be used frequently. Therefore, let's add it back!

### Use case

_No response_

### Additional Context

_No response_

roeldegoede:
## Issue addressed
Fixes #327 

## Explanation
The code that we used previously should still work with some minor adjustments. The only thing that is different is the way we store the upstream area and potential other attributes of the discharge points. Before, the srcfile could only contain the xy-coordinates, now we also allow a name, but still no data about upstream area. Therefore we used to write the additional information into a geojson file, which we also read while reading the srcfile. 

The netcdf format of the netsrcdisfile in theory allows for storing more information of each points, so maybe that is cleaner?

## Checklist

Still needs to be tested

### DIFF
--- a/hydromt_sfincs/components/forcing/boundary_conditions.py
+++ b/hydromt_sfincs/components/forcing/boundary_conditions.py
@@ -178,7 +178,7 @@ class SfincsBoundaryBase(ModelComponent):
                 gdf0 = gdf0_dedup
             # Create dataset with dummy values for new points and combine with existing dataset
             ds_new = self._create_dummy_dataset(gdf, ds0.time, value)
-            gds_new = GeoDataset.from_gdf(gdf, ds_new, keep_cols=False)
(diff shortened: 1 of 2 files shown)


vanasseltk (approved):
Mooi! Denk ik inderdaad goed om later een keer kritisch te kijken naar wat er thuis hoort in hydromt en hydromt_sfincs, kwa workflows.

vanasseltk (approved):

vanasseltk (approved):

PR #369, opened 2026-04-14
DDB-driven geometry/forcing adaptations

maartenvanormondt:
## Summary
Collection of small tweaks driven by DelftDashboard integration:
- Weirs / thin_dams / drainage_structures: consistent `z` column handling.
- Boundary conditions, snapwave boundary, wave makers: misc DDB-facing adjustments.
- `quadtree_elevation`: switch from `bathymetry_database.get_bathymetry_on_points` to `make_regular_grid + merge_multi_dataarrays` (data-catalog path).
- Defensive file-exists guards in several geometry / water_level readers.
- Observation points: return early when no `obsfile` is configured.

Note: the attribute-reading guard also touched `runup_gauges.py` in its original commit; that portion was dropped here because `runup_gauges.py` is delivered by PR #366 (runup gauges).

## Test plan
- [ ] Existing geometry / forcing tests pass
- [ ] Model build with a data-catalog elevation list (no bathymetry_database) works end to end

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/hydromt_sfincs/components/forcing/boundary_conditions.py
+++ b/hydromt_sfincs/components/forcing/boundary_conditions.py
@@ -182,7 +182,7 @@ class SfincsBoundaryBase(ModelComponent):
                 gdf0 = gdf0_dedup
             # Create dataset with dummy values for new points and combine with existing dataset
             ds_new = self._create_dummy_dataset(gdf, ds0.time, value)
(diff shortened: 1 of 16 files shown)


roeldegoede:
This PR throws away recently restructered code for quadtree_elevation. We should make sure it's up to date with main, and doesnt overwrite new stuff ...

PR #379, opened 2026-05-07
Feat/downscaling quadtree

keesnederhoff:
## Issue addressed
Add additional downscaling techniques

## Explanation
Downscaling: moved to separate module, kept raw and bilinear as options, and added 2 method-agnostic helper steps (dilation and energy head).
## Checklist

N/A

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,15 +23,7 @@ Added
 - platform aware launcher script (exe_path + write_batch_file) (#367)
 - New public helpers in ``hydromt_sfincs.workflows.downscaling``: ``dilate_zsmax`` (cell-space WSE dilation) and ``apply_energy_head`` (per-cell Bernoulli velocity-head correction). Both work on quadtree (``xu.UgridDataArray``) *and* regular (``xr.DataArray``) grids and preserve the wet-cell set.
 - ``downscale_floodmap`` gained two new kwargs: ``dilation`` (factor for the cell-space WSE dilation pre-step) and ``energy_flux`` (boolean switch for the Bernoulli velocity-head correction; requires ``qmax``). When ``energy_flux=True`` and ``method="bilinear"``, ``qmax`` is passed into ``_downscale_bilinear`` for per-cell Bernoulli + upstream-energy propagation (scaled by ``q_scale``); for any other method, :func:`apply_energy_head` runs as a pre-step (pure per-cell Bernoulli, no upstream propagation).
-
-Changed
--------
 - Reorganised the downscaling stack into the ``hydromt_sfincs.workflows.downscaling`` submodule. ``downscale_floodmap``, ``remove_disconnected_flooding``, and ``make_index_cog`` are now importable from ``hydromt_sfincs.workflows.downscaling`` (and re-exported via ``hydromt_sfincs.workflows``). ``hydromt_sfincs.utils`` now retains only generic I/O and grid utilities.
-
-Removed
--------
-- Removed the cell-based downscaling methods ``slope``, ``volume``, ``volume_slope``, and ``tilt_volume`` from ``downscale_floodmap``. Only ``raw``, ``constant``, and ``bilinear`` remain. These may be reinstated later if needed.
-- Dropped the ``sbg_fn`` kwarg from ``downscale_floodmap`` (only used by the removed cell-based methods).
 
 Fixed
 -----
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -40,9 +40,9 @@ Added
 - added back the option to create discharge forcing from a grid (previously named `setup_discharge_forcing_from_grid`) (#358)
 - add runup gauges geometry component (#366)
 - platform aware launcher script (exe_path + write_batch_file) (#367)
-- Reorganised the downscaling stack into the ``hydromt_sfincs.workflows.downscaling`` submodule with two method-agnostic WSE pre-steps — ``adjust_zsmax_dilation`` (cell-space WSE dilation) and ``adjust_zsmax_energyhead`` (per-cell Bernoulli velocity-head correction) — that work on quadtree (``xu.UgridDataArray``) *and* regular (``xr.DataArray``) grids and preserve the wet-cell set. ``downscale_floodmap``, ``remove_disconnected_flooding`` and ``make_index_cog`` are now importable from ``hydromt_sfincs.workflows.downscaling`` (and re-exported via ``hydromt_sfincs.workflows``); ``hydromt_sfincs.utils`` retains only generic I/O and grid utilities.
-- ``downscale_floodmap`` now supports ``raw``, ``constant`` and ``bilinear`` on both regular and quadtree grids. ``method`` is the single knob for the interpolation: on a regular grid ``constant`` uses nearest resampling (bathtub) and ``bilinear`` uses bilinear resampling (reproject engine); on a quadtree ``bilinear`` uses a scattered interpolator. ``raw`` paints the water level of the SFINCS cell that *contains* each pixel (exact containment via the index COG). All three methods use the index COG to speed up the computation (containment lookup for ``raw`` / ``constant``; empty-block skipping + edge masking for ``bilinear``).
-- ``downscale_floodmap`` returns the downscaled product as a (lazy) DataArray for *every* method — flood depth (``hmax``) for ``constant`` / ``bilinear`` and water level for ``raw`` — in addition to writing it to ``floodmap_fn`` / ``zsmap_fn``.
+- Moved the downscaling utilities from ``hydromt_sfincs.utils`` to a dedicated ``hydromt_sfincs.workflows.downscaling`` submodule (re-exported via ``hydromt_sfincs.workflows``).
+- ``downscale_floodmap`` is now parameterised by ``reproj_method`` (``nearest`` or ``bilinear``) and a ``subtract_dem`` flag (flood depth vs. raw water level), and works on both regular and quadtree grids.
+- Added reusable pre-processing (``adjust_zsmax_dilation``, ``adjust_zsmax_energyhead``) and post-processing (``remove_disconnected_flooding``) helpers as separate functions.
 
 Fixed
 -----
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,9 +21,9 @@ Added
 - added back the option to create discharge forcing from a grid (previously named `setup_discharge_forcing_from_grid`) (#358)
 - add runup gauges geometry component (#366)
 - platform aware launcher script (exe_path + write_batch_file) (#367)
-- New public helpers in ``hydromt_sfincs.workflows.downscaling``: ``dilate_zsmax`` (cell-space WSE dilation) and ``apply_energy_head`` (per-cell Bernoulli velocity-head correction). Both work on quadtree (``xu.UgridDataArray``) *and* regular (``xr.DataArray``) grids and preserve the wet-cell set.
-- ``downscale_floodmap`` gained two new kwargs: ``dilation`` (factor for the cell-space WSE dilation pre-step) and ``energy_flux`` (boolean switch for the Bernoulli velocity-head correction; requires ``qmax``). When ``energy_flux=True`` and ``method="bilinear"``, ``qmax`` is passed into ``_downscale_bilinear`` for per-cell Bernoulli + upstream-energy propagation (scaled by ``q_scale``); for any other method, :func:`apply_energy_head` runs as a pre-step (pure per-cell Bernoulli, no upstream propagation).
-- Reorganised the downscaling stack into the ``hydromt_sfincs.workflows.downscaling`` submodule. ``downscale_floodmap``, ``remove_disconnected_flooding``, and ``make_index_cog`` are now importable from ``hydromt_sfincs.workflows.downscaling`` (and re-exported via ``hydromt_sfincs.workflows``). ``hydromt_sfincs.utils`` now retains only generic I/O and grid utilities.
+- Reorganised the downscaling stack into the ``hydromt_sfincs.workflows.downscaling`` submodule with two method-agnostic WSE pre-steps — ``adjust_zsmax_dilation`` (cell-space WSE dilation) and ``adjust_zsmax_energyhead`` (per-cell Bernoulli velocity-head correction) — that work on quadtree (``xu.UgridDataArray``) *and* regular (``xr.DataArray``) grids and preserve the wet-cell set. ``downscale_floodmap``, ``remove_disconnected_flooding`` and ``make_index_cog`` are now importable from ``hydromt_sfincs.workflows.downscaling`` (and re-exported via ``hydromt_sfincs.workflows``); ``hydromt_sfincs.utils`` retains only generic I/O and grid utilities.
+- ``downscale_floodmap`` now supports ``raw``, ``constant`` and ``bilinear`` on both regular and quadtree grids. ``method`` is the single knob for the interpolation: on a regular grid ``constant`` uses nearest resampling (bathtub) and ``bilinear`` uses bilinear resampling (reproject engine); on a quadtree ``bilinear`` uses a scattered interpolator. ``raw`` paints the water level of the SFINCS cell that *contains* each pixel (exact containment via the index COG). All three methods use the index COG to speed up the computation (containment lookup for ``raw`` / ``constant``; empty-block skipping + edge masking for ``bilinear``).
+- ``downscale_floodmap`` returns the downscaled product as a (lazy) DataArray for *every* method — flood depth (``hmax``) for ``constant`` / ``bilinear`` and water level for ``raw`` — in addition to writing it to ``floodmap_fn`` / ``zsmap_fn``.
 
 Fixed
 -----
@@ -41,6 +41,11 @@ Changed
 - convert multipolygon into polygons and improve logging for creation of masks (#319)
 - some minor quadtree enhancements for missing and interpolating data (#352)
 - refactored config system: reorganized variables, improved keyword persistence logic, downgraded invalid keyword errors to warnings (#368)
+- ``downscale_floodmap`` is now a pure downscaler: the dilation and velocity-head adjustments are applied beforehand by calling ``adjust_zsmax_dilation`` / ``adjust_zsmax_energyhead`` on ``zsmax`` directly, instead of via the ``dilation`` / ``energy_flux`` kwargs.
+
+Removed
+-------
+- Removed the ``dilation``, ``energy_flux``, ``qmax``, ``zb``, ``q_threshold``, ``q_scale`` and ``reproj_method`` kwargs from ``downscale_floodmap`` (the resampling now follows from ``method``), together with the bilinear-only upstream-energy-propagation feature. The per-cell velocity head remains available via ``adjust_zsmax_energyhead``.
 
 v2.0.0-rc1 (25-11-2025)
 =============================
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,6 +21,17 @@ Added
 - added back the option to create discharge forcing from a grid (previously named `setup_discharge_forcing_from_grid`) (#358)
 - add runup gauges geometry component (#366)
 - platform aware launcher script (exe_path + write_batch_file) (#367)
+- New public helpers in ``hydromt_sfincs.workflows.downscaling``: ``dilate_zsmax`` (cell-space WSE dilation) and ``apply_energy_head`` (per-cell Bernoulli velocity-head correction). Both work on quadtree (``xu.UgridDataArray``) *and* regular (``xr.DataArray``) grids and preserve the wet-cell set.
+- ``downscale_floodmap`` gained two new kwargs: ``dilation`` (factor for the cell-space WSE dilation pre-step) and ``energy_flux`` (boolean switch for the Bernoulli velocity-head correction; requires ``qmax``). When ``energy_flux=True`` and ``method="bilinear"``, ``qmax`` is passed into ``_downscale_bilinear`` for per-cell Bernoulli + upstream-energy propagation (scaled by ``q_scale``); for any other method, :func:`apply_energy_head` runs as a pre-step (pure per-cell Bernoulli, no upstream propagation).
+
+Changed
+-------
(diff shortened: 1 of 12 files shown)


roeldegoede:
## Review — feat/downscaling-quadtree

Nice work overall! Moving the downscaling stuff to `workflows/downscaling.py` makes a lot of sense, that was long overdue, but I would move it on top of the file.  Adding `dilate_zsmax` and `apply_energy_head` as standalone pre-processing tools is a good idea — reusable and easy to test independently. Same for the post-processing helpers like     "compute_flow_connected_mask" and  "remove_disconnected_flooding",— good to have those separated out. 

A few things I'd like to see changed before merging:

- The downscale_floodmap method is still unreadable, but the signature exploded — from 10 to 17 parameters.which makes it even harder to use... The  `dilation`, `energy_flux`, `qmax`, `zb`, `q_threshold` — these are all preprocessing steps on `zsmax` that have nothing to do with the actual downscaling. Since we now have `dilate_zsmax` and `apply_energy_head` as standalone functions, the user should just call those themselves before calling `downscale_floodmap`. That keeps the main function clean and makes the preprocessing steps composable. 

- The `method` parameter is confusing and mostly redundant. We already had `reproj_method` which does the same thing. `method="constant"` is just `reproj_method="nearest"`, which we already had. `method="bilinear"` sounds similar to `reproj_method="linear"` (though that does only seem to work for regular grids in the old code). And `method="raw"` is the same as constant/nearest but without subtracting the DEM, so not a whole new method but easy to obtain with . Suggestion: drop `method`, keep `reproj_method`, add a `subtract_dem=True` flag if we want to get just water levels?

- The Bernoulli / energy-flux additions are too complex to have included in the main function? `apply_energy_head` and `_apply_upstream_propagation` are scientifically interesting ideas for high-velocity situations like steep rivers or weirs, but I don't really follow it yet.  More importantly, mixing these into `downscale_floodmap` makes the common case much harder to read. Since we already have `apply_energy_head` and `dilate_zsmax` as standalone functions, the cleaner pattern is to just let users call them before passing `zsmax` in — the orchestrator doesn't need to know about them. I'd suggest moving these out of the main function signature entirely and picking them up in a follow-up PR with proper tests and documentation.

- CRS handling was missing in the bilinear/raw fallback path with the new interpolation methods. Maybe this becomes unimportant when we go slightly back. When no index COG is provided the pixel coordinates need to be transformed to the model CRS before querying the interpolator — otherwise you get wrong results silently when model and DEM are in different projections. `make_index_cog` already handles this correctly with `Transformer.from_crs`, so the interpolation path should do the same.

- All other helper functions to make the files, overviews, masking make it more readable, so let's keep those!

hydromt_sfincs/components/forcing/boundary_conditions.py
```diff
@@ -137,7 +141,9 @@ def set(
             if index_dim != "index":
```
roeldegoede, line 141: Nice catch! I updated this a bit by moving the normalization into _ensure_canonical_index(). That way the internal canonical 0..N-1 index rule is explicit and shared across the different update paths, instead of being handled separetley all the time.

hydromt_sfincs/workflows/cog.py
```diff
@@ -9,7 +9,7 @@
 from rasterio.enums import Resampling
 from rasterio.transform import from_origin
 
-__all__ = ["make_index_cog", "make_topobathy_cog"]
+__all__ = ["make_quadtree_index_cog", "make_topobathy_cog"]
```
roeldegoede, line 13: Nice, the new names are more clear, but this functionality is a bit all over the place and double. Not sure if we should fix it now, but good to think about it.

I guess the most consistent would be:

**workflows/cog.py**, this is where the real workflows are described):
1. create_topopbathy_cog (which is independent on the grid actually, we just need an extent). This function does not work at the momemnt since it uses the bathymetry_database).
2. create_regular_index_cog
3. create_quadtree_index_cog

**worfklows/downscaling.py**, this is where we apply it
1. if we want to have the create_index_cog there, we could keep it like we have now, but let it call the create_regular or create_quadtree index_cog? Otherwise we have quite some double code

**quadtree.py and regulargrid.py**, model components
and then in the model componentens, SfincsGrid and SfincsQuadtreeGrid, we can actually also add/keep the methods create_index_cog and create_topobathy_cog, which then both would call the workflows from cog.py. Maybe a bit double, but it's quite nice that those methods would be available for scripted building based with the yaml configuraiton files.

I could pick this up, but wanted to discuss it first. We also have a lot of methods that do stuff with tiling, I guess there we should do the same (so workflows/tiling.py, and call those methods from the model components).

hydromt_sfincs/utils.py
```diff
@@ -259,8 +264,7 @@ def write_xyn(fn: str = "sfincs.obs", gdf: gpd.GeoDataFrame = None, fmt: str = "
 
     with open(fn, "w") as fid:
         for point in gdf.iterfeatures():
-            # only take first two coordinates if geometry is 3D
-            x, y = point["geometry"]["coordinates"][:2]
```
roeldegoede, line 263: Why was this needed? Actually some points (coming out of GIS for example) have a third dimension (X,Y,Z). For those points, this script was failing. The names are not part of the coordinates, so this was purely to drop any potential Z coordinate.

hydromt_sfincs/utils.py
```diff
@@ -230,8 +227,16 @@ def read_xy(fn: Union[str, Path], crs: Union[int, CRS] = None) -> gpd.GeoDataFra
...
+    # Read directly (only the x, y columns) so the parser is robust to a
+    # trailing name column, e.g. a sfincs.bnd written as ``x y "pointN"``.
+    # hydromt's open_vector(driver="xy") only renames columns 0/1 and then
+    # lower-cases all column names, which raises on the leftover integer name
+    # column.  Any columns beyond x, y are ignored here (use read_xyn for names).
+    df = pd.read_csv(fn, index_col=False, header=None, sep=r"\s+")
+    gdf = gpd.GeoDataFrame(geometry=gpd.points_from_xy(df[0], df[1]))
+    if crs is not None:
+        gdf.set_crs(crs, inplace=True)
+    gdf.index = np.arange(1, gdf.index.size + 1, dtype=int)  # index starts at 1
```
roeldegoede: Why would we need this? Maybe the existence of read_xy and read_xyn are a bit of an overkill, but read_xy should purely read-in x-y files, that we typically have for the bndfile etc. Th read_xyn is for the observation points and discharge_points. 

I also don't see why we have to change the index to 1 based, since before, in the boundary_conditions.py we made everything zero-based?

docs/changelog.rst
```diff
@@ -40,6 +40,9 @@ Added
 - added back the option to create discharge forcing from a grid (previously named `setup_discharge_forcing_from_grid`) (#358)
 - add runup gauges geometry component (#366)
 - platform aware launcher script (exe_path + write_batch_file) (#367)
+- Reorganised the downscaling stack into the ``hydromt_sfincs.workflows.downscaling`` submodule with two method-agnostic WSE pre-steps — ``adjust_zsmax_dilation`` (cell-space WSE dilation) and ``adjust_zsmax_energyhead`` (per-cell Bernoulli velocity-head correction) — that work on quadtree (``xu.UgridDataArray``) *and* regular (``xr.DataArray``) grids and preserve the wet-cell set. ``downscale_floodmap``, ``remove_disconnected_flooding`` and ``make_index_cog`` are now importable from ``hydromt_sfincs.workflows.downscaling`` (and re-exported via ``hydromt_sfincs.workflows``); ``hydromt_sfincs.utils`` retains only generic I/O and grid utilities.
```
roeldegoede: The changelog reads a bit like it is describing the evolution within this PR branch. I guess we could keep it much simpler here, e.g.:

“Moved downscaling utilities to hydromt_sfincs.workflows.downscaling”
“downscale_floodmap supports raw, constant, and bilinear”
“New pre-processing (dilate ...)  and post-processing (... ...) steps are now avaialble as separate helpers”

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -1,17 +1,70 @@
...
+#  Private helpers are prefixed with `_` and grouped under their step.
+# =============================================================================
+
+
+# =============================================================================
+#  STEP 0 — Setup: build the SFINCS cell-index COG (consumed by constant/raw)
+# =============================================================================
+
+
 def make_index_cog(
```
roeldegoede, line 68: I guess double code with cog.py. Could be nice to move the real stuff to cog.py, and keep i the general make_index_cog here

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+    * ``"raw"`` -- Paint each DEM pixel with the water level of the SFINCS
+      cell that *contains* it (the value straight from the SFINCS computation,
+      via exact containment in the index COG).  On a regular grid containment
+      is the nearest cell; on a quadtree it is not.  No DEM subtraction, no
+      wet/dry masking.  Produces a water-level raster.
+    * ``"constant"`` -- Assign each DEM pixel the WSE of its parent cell
+      (optionally via a pre-computed index COG), then subtract the DEM.
+      This is the classic "bathtub" approach.
+    * ``"bilinear"`` -- Bilinearly interpolate WSE from surrounding cell
+      centers (Sanders & Schubert 2019), then subtract the DEM.
```
roeldegoede: I don't fully see the added value of the "raw" method here as a separate method. I guess both for "constant" and "bilinear" it can be interesting to get the "raw" waterlevels as an ouput, so should we not simply make that an optional output for all methods when zsmax_fn is not None?

Then for the method "constant", I guess I prefer the name "nearest", and "bilinear" sounds good to me! I do see that in all cases it gets a bit funky when an index_cog is provided, since in that case, the index_cog already determines the interpolation method ... so it makes sense to only use that for the default method = "nearest"

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+            ):
+                raise ValueError("index should be str/Path when dep is str/Path.")
+        hmax = _downscale_floodmap_da(
+            zsmax=zsmax,
+            dep=dep,
+            indices=indices,
+            hmin=hmin,
+            gdf_mask=gdf_mask,
+            method=method,
+        )
```
roeldegoede: So this is where the real downscaling magic happens right, whether it's in-memory or file-based, we substract the bed levels from (interpolated) water levels. But to get the same resutls and no double code, we would also like to call this same method at the heart of the File-Based-Path.

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+        if zsmap_fn is None:
+            raise ValueError("zsmap_fn is required for method='raw'.")
+    else:
+        if floodmap_fn is None:
+            raise ValueError("floodmap_fn is required when dep is a file path.")
+
+    # Dispatch to the appropriate engine.  Regular-grid bilinear reuses the
+    # reproject engine (_downscale_constant, which derives the resampling from
+    # `method`); quadtree bilinear uses the scattered interpolator.
+    if method == "raw":
```
roeldegoede: So again, I don't see why we need these 3 different methods. For the file-based method, we do the same as above, but block-by-block. That logic is now in some nice helper functions, but the block -logic is the same for all methods, except the way we interpolate the water levels to the bed levels. If the _downscale_floodmap_da method also returns the interpolated water levels, we can write those to the file when zsmax_fn is defined?

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+    hmin,
+    gdf_mask,
+    floodmap_fn,
+    zsmap_fn,
+    method,
+    zoom_level,
+    nrmax,
+    logger,
+):
+    """File-based constant/bilinear downscaling via _downscale_floodmap_da.
```
roeldegoede: So I guess this one does contain everything we need, but we should apply all the helper function we created before. At least, the core function _downscale_floodmap_da is called.

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+                block_idx.raster.set_crs(indices_src.crs.to_wkt())
+
+            block_hmax = _downscale_floodmap_da(
+                zsmax=zsmax,
+                dep=block_dep,
+                indices=block_idx if indices is not None else None,
+                hmin=hmin,
+                gdf_mask=gdf_mask,
+                method=method,
+            )
```
roeldegoede: So here in theory, we could also return block_zsmax and write it to file when zsmax_fn is not None.

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+            # Mask the DEM's own nodata sentinel (e.g. -9999) to NaN.
+            if src.nodata is not None and not np.isnan(src.nodata):
+                dem_block[dem_block == src.nodata] = np.nan
+
+        xx, yy = _block_pixel_centres(geo["transform"], bm0, bm1, bn0, bn1)
+        zs_interp = (
+            interpolator(np.column_stack([xx.ravel(), yy.ravel()]))
+            .reshape(dem_block.shape)
+            .astype(np.float32)
+        )
```
roeldegoede: This interpolator part we could also use in _downscale_floodmap_da? I think we should fundamentelly test the performance of the different interpolators there on speed and accuracy of results. 

With the interpolators I mean, we have some from scipy that we use here and for nearest, we have some native ones in hydromt that we used for the regular grids, and we have some from xugrid ...

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+            inside = ~outside
+            if inside.any():
+                pidx = idx_block[inside].astype(int)
+                parent_zs = vals[pidx]
+                parent_H = H_eff[pidx]
+                dry_parent = np.isnan(parent_zs) & np.isnan(parent_H)
+                mask_arr = np.zeros_like(outside)
+                mask_arr[inside] = dry_parent
+                outside |= mask_arr
+            zs_interp[outside] = np.nan
```
roeldegoede: Nice! This is new functionality, which makes sense I guess. We could still have that in the core method, since we never want to add new wet cells?

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+    dep : Path, str, xr.DataArray
+        High-resolution DEM (m) of model region:
+    hmin : float, optional
+        Minimum water depth (m) to be considered as "flooded", by default 0.05
+    gdf_mask : gpd.GeoDataFrame, optional
+        Geodataframe with polygons to mask floodmap, example containing the landarea, by default None
+        Note that the area outside the polygons is set to nodata.
+    """
+
+    # Fast path: exact containing-cell lookup via the index COG.  Only valid
```
roeldegoede, line 1074: Okay so we either use the indices file, that makes sense I guess, and that already includes the "nearest" resampling methods, and eventually but not implemented yet other smart things. Good that we have warnings/errors that indices don't work for bilinear.

The other route we should verify whether we want to use the hydromt/xarray reprojections and xugrid reprojection, or whether the scipy ones are faster/better. But I guess we should just make sure that for regular water levels, and unstructured water levels, we are able to reproject it with "nearest" and "bilinear" methods.

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -132,3 +185,1509 @@ def make_index_cog(
...
+            idx_vals = np.squeeze(np.asarray(indices.values))
+            if idx_vals.shape != dep.shape:
+                raise ValueError(
+                    f"Indices shape {idx_vals.shape} does not match "
+                    f"dep shape {dep.shape}."
+                )
+            idx_nodata = indices.raster.nodata
+            if idx_nodata is None:
+                idx_nodata = 2147483647
+            hmax = hmax.where(idx_vals != idx_nodata)
```
roeldegoede: Allright nice, this is an important step I guess, which we could also do without the indices? Or will it become too time consuming then, since that would come down on also doing a nearest interpolation of the water levels.

hydromt_sfincs/workflows/downscaling.py
```diff
@@ -1,17 +1,70 @@
...
+#              1b. adjust_zsmax_energyhead        per-cell Bernoulli velocity head
+#    Step 2  Downscale
+#              2a. downscale_floodmap             public entry + per-method dispatch
+#              2b. helpers                        block streaming + output plumbing
+#              2c. engines                        raw / constant / bilinear + core
+#    Step 3  Post-process the downscaled raster (optional)
+#              3a. remove_disconnected_flooding   drop disconnected wet pools
+#              3b. compute_flow_connected_mask    cell-level connectivity helper
+#
+#  Private helpers are prefixed with `_` and grouped under their step.
```
roeldegoede, line 59: This is really nice! Haven't tested the pre- and post-steps extensively, but I like the idea and structure!

hydromt_sfincs/workflows/cog.py
roeldegoede, line 21: Good to realize that this functionality does not work. (Not your fault ofcourse)

roeldegoede (changes requested):
Overall really nicely restructred and big improvement! There were a couple of things changed as well related to 0-indexing in boundary_conditions and the utils which I coulnd't completely follow, and there is still quite some double code in the downscaling routines.

docs/changelog.rst
keesnederhoff: Done, trimmed to three lines: downscaling moved to `workflows.downscaling`, `downscale_floodmap` parameterised by `reproj_method` + `subtract_dem`, and the pre/post-processing helpers listed separately.

hydromt_sfincs/components/forcing/boundary_conditions.py
keesnederhoff: Agreed, that's cleaner.

hydromt_sfincs/utils.py
keesnederhoff: You're right, that was inconsistent. Put it back to 0-based so it matches the canonical index in `boundary_conditions`. `read_xy` now just reads the x/y columns, names stay in `read_xyn`.

hydromt_sfincs/utils.py
keesnederhoff: Exactly, the `[:2]` drops a possible Z on 3D points

hydromt_sfincs/workflows/cog.py
keesnederhoff: Good point, and I agree with the split. Shall we keep this out of the PR and pick it up in a separate one?

hydromt_sfincs/workflows/cog.py
keesnederhoff: Yes, let's fix it as part of the cog.py!

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Thanks!

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Yes, should be addressed in a cog.py reorganisation

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Done as you suggested. `method` is gone. `raw` is just `subtract_dem=False` (raw water level, for either interpolation), `constant` is `reproj_method="nearest"`, `bilinear` is `reproj_method="bilinear"`. So the raw water level is available as an output for any method, and the index-COG fast path is used for `nearest` only (bilinear uses the index just to mask).

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Yes, agree

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Fair. Reverted to just the 2.

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Yep. `_downscale_raw` and `_downscale_bilinear` are gone; everything runs through the shared block loop and `_downscale_floodmap_da`.

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Yes, done

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: The refactor puts the backend in one place (`_interp_zs_on_dep`): regular via hydromt `reproject_like`, quadtree via a scipy scattered interpolant built once and reused across blocks. I haven't done a proper speed benchmark.

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: Yes indeed!

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: The routes are explicit now: with an index COG it's exact containment (nearest only); without one it's hydromt reproject (regular) or scipy scattered (quadtree). Both `nearest` and `bilinear` work on regular and quadtree. For `bilinear` the index is used only to mask, with a warning.

hydromt_sfincs/workflows/downscaling.py
keesnederhoff: It splits by route. For nearest, it's already free: the containment step that does the downscaling (`rasterize_like` on a quadtree, `reproject_like` on a regular grid) returns nodata outside the domain and over dry cells, so the mask falls out for free. That's why `constant_noidx` matched `constant_idx` byte-for-byte. For bilinear, it costs a second pass. The linear interpolant only knows the convex hull of the wet cell centres, so without an index it bleeds across interior dry-cell gaps: on the full Snohomish test run I did that's 2.5% wet with the index vs 6.3% without. To mask that, without an index, you'd need a separate nearest/`rasterize_like` pass just for the mask. So it's doable, but the index does both jobs cheaply with one precomputed raster. I'd keep it as the recommended path, with the convex-hull fallback (plus the warning we already emit) for the no-index case.

hydromt_sfincs/workflows/cog.py
roeldegoede: Yes, that's fine!

roeldegoede (approved):

PR #387, opened 2026-06-17
Feat: separate function for reading config

dalmijn:
## Issue addressed
Fixes nothing in particular

## Explanation
Separated the reading/ parsing part of the `read` method of the config component into a separate function in the readers submodule. I think it can be nice to be able to read the config file outside of the context of the component/ model. This doesn't need to be in `readers.py` but that was just my first guess. Feel free to close it, if you fully disagree with this PR.
EDIT: Shall I just put in in `utils` or might something like the `readers` and `writers` modules such as in core be an idea.

## Checklist

### DIFF
--- a/docs/api/api.rst
+++ b/docs/api/api.rst
@@ -239,6 +239,7 @@ Input/Output methods
 .. autosummary::
    :toctree: ../_generated/
 
+   readers.read_config
    utils.read_binary_map
    utils.write_binary_map
    utils.read_binary_map_index
--- a/hydromt_sfincs/components/config/config.py
+++ b/hydromt_sfincs/components/config/config.py
@@ -1,7 +1,5 @@
 import logging
-from ast import literal_eval
-from datetime import datetime
-from os.path import abspath, exists, isabs, join
+from os.path import abspath, isabs, join
 from pathlib import Path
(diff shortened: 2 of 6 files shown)


roeldegoede:
Nice, I like the way of thinking, and not much to add to the code! But before merging, it might be a good moment actually to move all readers from the utils script to this separate readers.py before we make the final v2.0.0 release? I see there is a readers and writers script in hydromt-core, is that also the way other plugins deal with it? What do you think?

dalmijn:
> Nice, I like the way of thinking, and not much to add to the code! But before merging, it might be a good moment actually to move all readers from the utils script to this separate readers.py before we make the final v2.0.0 release? I see there is a readers and writers script in hydromt-core, is that also the way other plugins deal with it? What do you think?

For the few I/O function in hydromt-fiat I will, I indeed like the way it is done like that in core, and I see hydromt-sfincs has quite a few read and write functions on it's own. I'll happily separate the utils I/O functions into the reader and writer submodules in this PR or a new one (which might be clearer). Would that be okay with you?

roeldegoede:
Awesome, that sounds good! So the we can merge this one first, and then separate the other read/write functions in a new PR!

roeldegoede (approved):
Nice work!

Issue #394, opened 2026-07-07
plot_basemap colormap creation still fails for flat low-lying and below-sea-level models (follow-up to #324)


T-Sadana:
### HydroMT-SFINCS version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT-SFINCS.

### Reproducible Example

import numpy as np, xarray as xr, hydromt
from hydromt_sfincs.plots import plot_basemap
ds = xr.DataArray(np.random.uniform(0.2, 0.7, (20, 20)), coords={"y": np.linspace(50, 50.01, 20), "x": np.linspace(5, 5.01, 20)}, name="dep").to_dataset()
ds.raster.set_crs(4326); ds["dep"].raster.set_nodata(-9999.0)
plot_basemap(ds, geoms={}, variable="dep", plot_bounds=False, plot_region=False, plot_geoms=False)

### Current behaviour

Follow-up to #324 (also #325): `plot_basemap` can still fail to create the custom
terrain colormap for `dep`, for models in flat low-lying (coastal) areas.

The `vmin == vmax` guard added in #325 compares the **float** quantile values, but the
`int()` truncation happens *after* it, so `vmin`/`vmax` can still collapse to values for
which no valid colormap can be built:

https://github.com/Deltares/hydromt_sfincs/blob/82c101a8e86c439b83a7b2053323b0b069f24207/hydromt_sfincs/plots.py

https://github.com/Deltares/hydromt_sfincs/blob/82c101a8e86c439b83a7b2053323b0b069f24207/hydromt_sfincs/plots.py#L280-L292

### Desired behaviour

What fixed it for all our crashing models: apply the guard after the truncation instead, and give the colormap at least two samples, i.e. replace lines 283-287 with

vmin, vmax = int(kwargs.pop("vmin", vmin)), int(kwargs.pop("vmax", vmax))
if vmax <= vmin:
    vmax = vmin + 1
c_dem = plt.cm.terrain(np.linspace(0.25, 1, max(2, vmax)))
Normal-terrain plots are unchanged by this.

### Additional context

[...]

PR #395, opened 2026-07-08
Fix/plots and elevation.create()

### LINKED ISSUE #392: Only interpolate nan values in elevation data within masked area
### Kind of request

Adding new functionality

### Enhancement Description

It would be helpful to add an extra optional mask argument to `elevation.create` with a polygon. This mask could then be used to determine if there is a need to interpolate nan data within the area of interest. Especially for large areas with complicated masks, this could save a lot of computation time. See example in Mozambique flood plains below.

Link to relevant code:

https://github.com/Deltares/hydromt_sfincs/blob/fc8462ef12534dde3efe22dc0bec36c1ce3d6ff1/hydromt_sfincs/components/grid/elevation.py#L104

Additionally, being able to raise an error for missing data instead of interpolating would be great.

Image

### Use case

_No response_

### Additional Context

_No response_

roeldegoede:
## Issue addressed
Fixes #392 and #394

### DIFF
--- a/hydromt_sfincs/components/grid/elevation.py
+++ b/hydromt_sfincs/components/grid/elevation.py
@@ -65,6 +65,7 @@ class SfincsElevation(ModelComponent):
         elevation_list: List[dict],
         buffer_cells: int = 0,  # not in list
         interp_method: str = "linear",  # used for buffer cells only
+        interpolate_na: bool = True,
     ):
         """Interpolate topobathy (dep) data to the model grid.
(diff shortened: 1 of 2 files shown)


roeldegoede:
@jensdebruijn and @T-Sadana, could you check whether these changes are sufficient for your issues #392 and #394?

jensdebruijn:
For #392 this works perfectly, thanks!

PR #398, opened 2026-07-21
chaning the defaults of the manning values into None, so that when th…

### LINKED ISSUE #396: config.write() drops manning_land/manning_sea/rgh_lev_land when they equal the pydantic default, which mismatches SFINCS's -999 sentinel
### Describe the bug

`SfincsConfig.write()` omits any field whose value equals its pydantic default (unless the field is tagged `always`). For `manning_land`, `manning_sea` and `rgh_lev_land` those pydantic defaults are `0.04`, `0.02` and `0.0`, but the corresponding **SFINCS** defaults are the sentinel `-999.0` meaning "not set".

The result: a user who explicitly asks for the canonical land/sea Manning split gets those lines silently dropped from `sfincs.inp`. SFINCS then hits its `-999` fallback and applies a **spatially uniform** `manning` (0.04) to land *and* sea, so the requested split is lost without any warning. On a shallow shelf this doubles the intended bed friction and visibly damps wind-driven surge.

The values round-trip correctly through `update()`; the loss happens only at `write()`.

### Reproduction

```python
from hydromt_sfincs import SfincsModel

for tag, vals in [
    ("A", {"manning_sea": 0.02, "manning_land": 0.04, "rgh_lev_land": 0.0}),   # == hydromt defaults
    ("B", {"manning_sea": 0.03, "manning_land": 0.05, "rgh_lev_land": 0.1}),   # != hydromt defaults
]:
    m = SfincsModel(root=f"m_{tag}", mode="w+")
    m.config.update(vals)
    print(tag, "config.data:", {k: getattr(m.config.data, k) for k in vals})
    m.config.write()
    print(tag, "sfincs.inp:", [l.strip() for l in open(f"m_{tag}/sfincs.inp") if "manning" in l or "rgh_lev" in l])
```

Output:

```
A config.data: {'manning_sea': 0.02, 'manning_land': 0.04, 'rgh_lev_land': 0.0}
A sfincs.inp: []                                        <-- silently dropped
B config.data: {'manning_sea': 0.03, 'manning_land': 0.05, 'rgh_lev_land': 0.1}
B sfincs.inp: ['manning_land         = 0.05', 'manning_sea          = 0.03', 'rgh_lev_land         = 0.1']
```

Case A is the common one: 0.02 over water and 0.04 over land are the values most people want.

### Relevant code

`hydromt_sfincs/components/config/config.py`, in `write()`:

```python
always = extra.get("always", False)
if not always:
    ...
    if value == field_info.default:
        continue        # not written
```

`hydromt_sfincs/components/config/config_variables.py`:

```python
manning_land: float = Field(0.04, description="...; -999 = not set ...")
manning_sea:  float = Field(0.02, description="...; -999 = not set ...")
rgh_lev_land: float = Field(0.0,  description="...")
```

Neither field carries `json_schema_extra={"always": True}` (unlike e.g. `advection`).

SFINCS side, `sfincs_domain.f90:2026-2036`:

```fortran
if (manning_land<0.0) then
   gn2land = 9.81*manning**2      ! falls back to uniform manning
if (manning_sea<0.0) then
   gn2sea  = 9.81*manning**2
```

with `manning=0.04`, `manning_land=-999.0`, `manning_sea=-999.0` at `sfincs_input.f90:73-76`.

Note that `manning` in `sfincs_map.nc` is all-NaN in my runs, so the mistake is not visible in the output either; the only way to see it is to read `sfincs.inp`.

### Suggested fix

Either:

- set the pydantic defaults to `-999.0` for `manning_land` / `manning_sea`, matching the SFINCS sentinel, so an explicitly requested `0.02` is no longer "equal to the default"; or
- tag these three fields `json_schema_extra={"always": True}` so they are always emitted when set.

The first seems more correct, since it makes the hydromt default mean the same thing as the SFINCS default. It would also make the general rule safe again: "skip if equal to default" is only sound when the two layers agree on what the default means.

### Version

`hydromt_sfincs` 2.0.0-rc3; the same `write()` logic is present on `main`. SFINCS kernel `v2.4.0_Galibier_release`.

roeldegoede:
…ey get set, they are actually also written to the config

## Issue addressed
Fixes #396 

## Explanation
I think that with this update, we can now make a distinction between the default and recommended values. With the default to None, we guarantee that whenever the values are set to the recommended values of 0.02 resp 0.04, those new values will be written. When they are not set, they remain None, and they are not written to the sfincs.inp file

## Checklist


vanasseltk (commented):
So only default value for spatially constant manning? Manning_land and manning_sea can be set seperately

vanasseltk (approved):

Issue #173, opened 2024-01-05
Create geojson of snapped crosssections


TBovenschen:
### Kind of request

Adding new functionality

### Enhancement Description

Crosssections are being snapped to the SFINCS grid, however these snapped crosssections are not written as output. Therefore it is hard to know how your crosssections look exactly within the SFINCS model. You could derive this from overlaying the drawn crosssections with the grid, but I think it would be an easy and useful step to produce a geojson of the snapped crosssections.

### Use case

This could be useful if you want to know whether your snapped crosssections are crossing the entire floodplain. If not, water could flow around the crosssections.

roeldegoede:
I don't want to include this by default, since grid_snapping can be a computational intensive step. However, with the recent changes of #400, it should be very easy to snap them yourselves and inspect the outcomes. 

For example:
```
# snapping any geodataframe with linestring to the grid
gdf_any = geopandas.read_file( ...) 
gdf_snap = SfincsModel.quadtree_grid.snap_to_grid(gdf_any)

# or snapping specifically the cross_sections (or weirs etc)
gdf_snap = SfincsModel.cross_sections.snap_to_grid()
gdf_snap.to_file(cross_sections_snapped.geojson, driver="GeoJSON")
```

PR #400, opened 2026-07-22
Enh/grid snapping regular

roeldegoede:
## Issue addressed
Fixes snapping geometries to regular grids. This is key for #173, though it's still something the user has to call manually.

## Explanation
We now convert to regular grids into quadtree grids the first time quadtree_grid.data is called by any of the snap_to_grid functionality. Before we did it when reading in the regular models, but this way, we can also use the snap_to_grid stuff when we build a model from scratch (so read is not triggered).

## Checklist


panosatha (approved):
LGTM!

PR #402, opened 2026-07-22
provide clear warning for when the netcdf meteo forcing is not coveri…

### LINKED ISSUE #322: Add warning/error when netcdf meteo is not covering full model period
### Kind of request

None

### Enhancement Description

The SFINCS kernel seems to crash with unclear warnings when for example the netcdf meteo forcing (e.g. precipitation) starts later then tstart. I think we should make sure in the Meteo classes that this does not happen.

### Use case

_No response_

### Additional Context

_No response_

roeldegoede:
…ng the full model period --> especially tstart causes the kernel to crash

## Issue addressed
Fixes #322

## Explanation
Now we get a value error when the data that we finally set to the model does not cover the start, and a warning when the timeseries are too short.

## Checklist

### DIFF
--- a/hydromt_sfincs/components/forcing/meteo.py
+++ b/hydromt_sfincs/components/forcing/meteo.py
@@ -237,18 +237,11 @@ class SfincsMeteo(ModelComponent):
         # Get the start and end time of the data
         time_start = data.indexes["time"][0].to_pydatetime()
         time_end = data.indexes["time"][-1].to_pydatetime()
-        if time_start > model_start:
-            raise ValueError(
-                f"{name} forcing starts after the model start time. "
-                f"Model starts at {model_start}, {name} starts at "
-                f"{time_start}."
-            )
-
-        if time_end < model_end:
+        if time_start > model_start or time_end < model_end:
             logger.warning(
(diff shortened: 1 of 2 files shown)


hydromt_sfincs/components/forcing/meteo.py
```diff
@@ -232,6 +232,25 @@ def set(
...
+        model_start, model_end = self.model.get_model_time()
+        # Get the start and end time of the data
+        time_start = data.indexes["time"][0].to_pydatetime()
+        time_end = data.indexes["time"][-1].to_pydatetime()
+        if time_start > model_start:
+            raise ValueError(
+                f"{name} forcing starts after the model start time. "
+                f"Model starts at {model_start}, {name} starts at "
+                f"{time_start}."
+            )
```
vanasseltk: Why should this lead to an error? If I supply a meteodataset which starts after model time start, does SFINCS crash? Or does is assume zeros (no forcing)?

vanasseltk (approved):

vanasseltk (approved):

vanasseltk (approved):

PR #403, opened 2026-07-22
clean-up tiling.py, cog.py and downscaling.py

### LINKED ISSUE #399: Clean-up cog, tiling and downscale workflows.
This functionality for downscaling floodmaps is a bit all over the place and double. 

I guess the most consistent would be:

**workflows/cog.py**, this is where the real workflows are described):

1. create_topopbathy_cog (which is independent on the grid actually, we just need an extent). This function does not work at the momemnt since it uses the bathymetry_database).
2. create_regular_index_cog
3. create_quadtree_index_cog

**worfklows/downscaling.py**, this is where we apply it

if we want to have the create_index_cog there, we could keep it like we have now, but let it call the create_regular or create_quadtree index_cog? Otherwise we have quite some double code

**quadtree.py **and** regulargrid.py**, model components
and then in the model componentens, SfincsGrid and SfincsQuadtreeGrid, we can actually also add/keep the methods create_index_cog and create_topobathy_cog, which then both would call the workflows from cog.py. Maybe a bit double, but it's quite nice that those methods would be available for scripted building based with the yaml configuraiton files.

I could pick this up, but wanted to discuss it first. We also have a lot of methods that do stuff with tiling, I guess there we should do the same (so workflows/tiling.py, and call those methods from the model components).

roeldegoede:
## Issue addressed
Fixes #399 

## Checklist

### DIFF
--- a/docs/api/api.rst
+++ b/docs/api/api.rst
@@ -226,7 +226,6 @@ SFINCS workflows
    workflows.river_centerline_from_hydrography
    workflows.landuse
    workflows.cn_to_s
-   workflows.create_topobathy_tiles
 
 Flood map downscaling
 ---------------------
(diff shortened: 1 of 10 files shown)


roeldegoede:
@keesnederhoff I made a start with moving around the functionality related to the tiling and cloud-optimized-geotiffs. What do you think?

## Done so far
- [x] Moved make_index_cog to cog.py (and renamed to create_index_cog)
- [x] Added create_topobathy_cog (based on subgrid block approach to overcome memory issues)
- [x] Both index and topobathy cog now work for regular and quadtree models 
- [x] Moved all tiling logic to tiling.py and working for regular and quadtree models
- [x] Made regulargrid.py and quadtree.py consistent in calling all methods from cog.py and tiling.py

## To do
- [ ] Add some tests for index cog of regular grids, and topobathy_cog
- [ ] Probably some tests for tiling.py wouldn't hurt either
- [ ] Check DDB specific logic of create_topobathy_tiles ...
- [ ] Check if downscale_floodmap_webmercator is still needed/used

PR #404, opened 2026-07-23
Remove unnecessary precision while writing subgrid tables

roeldegoede:
np.zeros() defaults to float64, so each of these silently allocates twice

the memory actually needed

## Issue addressed
Every subgrid array built in SubgridTable.create() is explicitly allocated
as float32 (z_zmin, z_zmax, z_volmax, z_level, u_zmin/u_zmax/u_havg/u_nrep/
u_pwet/u_ffit/u_navg, and the v_* equivalents -- confirmed by reading every
np.full(..., dtype=np.float32) call in create(), lines ~139-174 and
~738-770 as of this writing).

write_netcdf() (lines ~382-449) re-derives active-cell-only versions of
these same arrays before writing them to the output .nc file, via three
np.zeros(...) calls that do NOT specify a dtype:

    z_level = np.zeros((nr_levels, nr_z_points))                  # line ~416
    uv_var  = np.zeros(nr_uv_points)                               # line ~424
    uv_var  = np.zeros((nr_levels, nr_uv_points))                  # line ~435

np.zeros() defaults to float64, so each of these silently allocates twice
the memory actually needed -- for data that is float32-precision on both
sides of the assignment (the right-hand side, ds["u_" + var].values, is
always float32). The extra precision is never used: the float32 source
values are simply upcast into a float64 buffer and later written to netCDF,
gaining nothing. For large domains at fine resolution this routinely adds
1-3 GiB of avoidable peak memory at exactly the step where our builds were
already memory-constrained, and was the direct cause of the crash quoted
above (nr_uv_points=12,438,241 * 30 levels * 8 bytes = 2.78 GiB; at float32
this becomes 1.39 GiB).

## Explanation
Add dtype=np.float32 to all three allocations, matching create()'s own
dtype convention. This is a pure memory-efficiency fix with zero effect on
numerical results (see verification below): every value written is already
float32-precision before this function ever runs.

## Checklist

### DIFF
--- a/hydromt_sfincs/components/grid/subgrid.py
+++ b/hydromt_sfincs/components/grid/subgrid.py
@@ -413,7 +413,7 @@ class SfincsSubgridTable(ModelComponent):
                 ds[var].values.flatten()[active_cells], dims=("np")
             )
 
-        z_level = np.zeros((nr_levels, nr_z_points))
+        z_level = np.zeros((nr_levels, nr_z_points), dtype=np.float32)
         for ilevel in range(nr_levels):
             z_level[ilevel] = ds["z_level"][ilevel].values.flatten()[active_cells]
         ds_new["z_level"] = xr.DataArray(z_level, dims=("levels", "np"))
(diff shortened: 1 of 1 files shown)


vanasseltk (commented):
Seems like a good choice as, why "is ds["u_" + var].values, is always float32" ? Maybe for later we need and option where you could change form 32 to 64 (is wanted by the user). 32 is a good default for now.

vanasseltk (approved):

PR #409, opened 2026-08-13
Fix SnapWave quadtree mask model kwarg handling

Leynse:
## Summary
- Align SnapWaveQuadtreeMask overridden method signatures with model kwarg forwarding
- Validate that only model='snapwave' is accepted in SnapWave-specific quadtree mask methods
- Add regression coverage for quadtree_snapwave_mask.create_active(model=...) calls

## Validation
- Ran targeted snapwave boundary tests (2 passed)

### DIFF
--- a/hydromt_sfincs/components/quadtree/snapwave_quadtree_mask.py
+++ b/hydromt_sfincs/components/quadtree/snapwave_quadtree_mask.py
@@ -41,9 +41,15 @@ class SnapWaveQuadtreeMask(SfincsQuadtreeMask):
         mask = self.data["snapwave_mask"]
         return (mask == 2).any().item()
 
+    @staticmethod
+    def _validate_model(model: str) -> None:
+        if model != "snapwave":
+            raise ValueError("SnapWaveQuadtreeMask only supports model='snapwave'.")
+
     @hydromt_step
     def create(
         self,
+        model: str = "snapwave",
         zmin: float = None,
         zmax: float = None,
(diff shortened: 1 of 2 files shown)


roeldegoede:
I don't really see why this is needed? When calling SfincModel.snapwave_quadtree_mask.create /create_active /create_boundary, the model should alwas be "snapwave" right? So why would we make this something the user can change?

Leynse:
That is what I thought, but still it didn't work in DelftDashboard as is.
It's something related to 'super', Copilot:

Fix it in hydromt_sfincs by making the SnapWave override signatures match the base quadtree mask API.

In your checked tree, the mismatch is in snapwave_quadtree_mask.py:

create_active does not accept model
but it forwards model="snapwave" to super().create_active(...) at line 156
Recommended fix (robust)
In snapwave_quadtree_mask.py, add model: str = "snapwave" to overridden methods (at least create_active, and ideally also create / create_boundary) so calls from base with model= won’t fail.

Also enforce SnapWave-only usage, e.g. raise if model != "snapwave".

If your parent quadtree class in quadtree_mask.py does not support model, then remove model=... from the super() calls instead.

So the key rule is: base and override method signatures + forwarded kwargs must be consistent

Leynse:
If you know a better way to fix in DelftDashboard itself, that could also work.
But this seemed easiest to fix "So the key rule is: base and override method signatures + forwarded kwargs must be consistent."

roeldegoede (approved):
Okay, well it doesn't hurt anyone this way, so let's merge!

